PR #20, opened 2026-05-05
Create a Django project with basic applications

MykytaKa:
Initialized the base Django project structure. Created and registered core apps (accounts, drones, roles) and configured basic URL routing for each application, linking them to the main project URLs.


config/settings.py
```diff
@@ -0,0 +1,124 @@
...
+
+# Load .env variables ]
+dotenv_path = BASE_DIR / '.env'
+load_dotenv(dotenv_path)
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/6.0/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = os.getenv('DJANGO_SECRET_KEY')
```
mehalyna, line 27: `SECRET_KEY` can become `None`, which prevents Django from starting securely and can fail unpredictably at runtime. Validate the environment variable at startup and raise a clear error if it is missing.
```
SECRET_KEY = os.getenv("DJANGO_SECRET_KEY")
if not SECRET_KEY:
    raise RuntimeError("DJANGO_SECRET_KEY is required")

```

.env.example
mehalyna, line 1: Add variables, eg.
```
DJANGO_SECRET_KEY=change-me
DEBUG=False
ALLOWED_HOSTS=localhost,127.0.0.1
```

mehalyna (approved):

.idea/modules.xml
torii044, line 1: Shouldn't these files be in .gitignore?

torii044 (approved):

yedynetsy (changes requested):
I guess .idea folder shouldn't be in the commit. Also, it seems like there is a missing root .gitignore file, where the .idea should be included.

yedynetsy (approved):

PR #43, opened 2026-05-09
Add user profile fields and admin support

M1ro1:
Done
User, Role, and MilitaryUnit models with relations (role, unit, created_by)
Admin registration for User, Role, MilitaryUnit, UserProfile
POST endpoint at /accounts/users/ for user creation
Password hashing via set_password and initial credentials email
UserProfile for rank, contact, profile_picture

Questions
Confirm whether audit logging is required (currently removed)
Decide on /api/users/ vs /accounts/users/ as the final route


accounts/serializers.py
anastasiiahumeniuk, line 1: the serializer currently depends on the model unique constraints for username and email. thats okay, but it would be nicer if the serializer returned a clean validation error before save instead of relying on the DB layer. Maybe add  some methods with text error like "Username already exists" and Email already exists"

anastasiiahumeniuk (approved):

anastasiiahumeniuk (approved):

accounts/serializers.py
```diff
@@ -0,0 +1,63 @@
...
+            'contact',
+            'profile_picture',
+        )
+        read_only_fields = ('id',)
+
+    @staticmethod
+    def validate_username(self, value):
+        if User.objects.filter(username__iexact=value).exists():
+            raise serializers.ValidationError("A user with that username already exists.")
+        return value
```
mehalyna: The validator is declared with the wrong signature and @staticmethod, so DRF will not call it correctly. Make it an instance method with (self, value) or a plain method without @staticmethod.

accounts/serializers.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @staticmethod
+    def validate_email(self, value):
+        if User.objects.filter(email__iexact=value).exists():
+            raise serializers.ValidationError("A user with that email already exists.")
+        return value
```
mehalyna: The email validator has the same broken signature and will be skipped or fail unexpectedly. Remove `@staticmethod` and keep the standard DRF validator shape. 
```
def validate_email(self, value):
    if User.objects.filter(email__iexact=value).exists():
        raise serializers.ValidationError("A user with that email already exists.")
    return value
```

config/settings.py
```diff
@@ -107,6 +108,15 @@
     },
 ]
 
+AUTH_USER_MODEL = 'accounts.User'
```
mehalyna: Introducing a custom user model in a non-initial migration path is a high-risk auth break because Django requires this before the first auth migration in a project lifecycle. Confirm this is a fresh database only or replace with a one-to-one profile approach for existing projects.

accounts/migrations/0001_initial.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        ('auth', '0012_alter_user_first_name_max_length'),
+        ('roles', '0001_initial'),
+    ]
```
mehalyna, line 18: This initial migration depends on `auth`, while the same PR swaps `AUTH_USER_MODEL` to `accounts.User`, which can create an invalid migration graph or startup failure. Make the custom user model the first app-level auth model decision and regenerate migrations from a clean state.

accounts/services.py
```diff
@@ -0,0 +1,23 @@
...
+    return get_random_string(length=12)
+
+def send_activation_email(user, password: str) -> None:
+    subject = "Your Account Has Been Created"
+    message = (
+        f"Welcome, {user.first_name}!\n\n"
+        f"Your account in the system has been created..\n"
+        f"Login: {user.username}\n"
+        f"Temporary password: {password}\n\n"
+    )
```
mehalyna, line 51: Sending a plaintext temporary password by email is a security anti-pattern. Send a password-set or activation link instead of credentials. 
```
token = default_token_generator.make_token(user) 
url = f"{settings.FRONTEND_URL}/activate/{user.pk}/{token}" 
send_mail(subject, f"Set your password: {url}", settings.DEFAULT_FROM_EMAIL, [user.email])
```

accounts/serializers.py
```diff
@@ -0,0 +1,63 @@
...
+            UserProfile.objects.create(
+                user=user,
+                rank=rank,
+                contact=contact,
+                profile_picture=profile_picture,
+            )
+
+        send_activation_email(user, password)
+
+        return user
```
mehalyna: The serializer mixes validation, persistence, password generation, profile creation, and email delivery, which is undue coupling and an SRP violation. Move orchestration into a dedicated service and keep the serializer thin.

accounts/views.py
```diff
@@ -1,3 +1,10 @@
...
 
-# Create your views here.
+from .models import User
+from .serializers import UserRegistrationSerializer
+from .permissions import IsSystemAdmin
+
+class UserRegistrationView(generics.CreateAPIView):
+    queryset = User.objects.all()
+    serializer_class = UserRegistrationSerializer
+    permission_classes = [IsSystemAdmin]
```
mehalyna: The view exposes full user creation without any transaction or explicit failure handling, so email errors can leave partially created users. Wrap creation in an atomic transaction or defer email sending after commit. 
```
from django.db import transaction

@transaction.atomic 
def perform_create(self, serializer): 
    serializer.save()
```

config/urls.py
```diff
@@ -20,6 +20,7 @@
 urlpatterns = [
     path('admin/', admin.site.urls),
     path('accounts/', include('accounts.urls')),
+    path('api/', include('accounts.urls')),
     path('drones/', include('drones.urls')),
     path('roles/', include('roles.urls')),
 ]
```
mehalyna, line 26: Mounting the same `accounts.urls` under both `/accounts/` and `/api/` creates duplicate endpoints and unclear API boundaries. Keep one canonical route, preferably under `/api/`.
```
urlpatterns = [
    path("admin/", admin.site.urls),
    path("api/accounts/", include("accounts.urls")),
]
```

accounts/permissions.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.permissions import BasePermission
+
+class IsSystemAdmin(BasePermission):
+    def has_permission(self, request, view):
+        if not request.user or not request.user.is_authenticated:
+            return False
+
+        role = getattr(request.user, "role", None)
+        return bool(role and role.code == "ADMIN")
```
mehalyna: Hard-coding "ADMIN" couples permissions to a magic string and spreads role policy into low-level request checks. Centralize role codes in a constant or model enum and reuse it here. 
```
from roles.models import ROLE_CODES

ADMIN_CODE = "ADMIN" return bool(role and role.code == ADMIN_CODE)
```

accounts/admin.py
```diff
@@ -1,3 +1,40 @@
...
+                "fields": (
+                    "role",
+                    "unit",
+                    "created_by",
+                    "created_at",
+                    "updated_at",
+                )
+            },
+        ),
+    )
```
mehalyna, line 23: Appending custom fields to `UserAdmin.fieldsets` without `add_fieldsets` means the admin add form will not expose the new required fields consistently. Define `add_fieldsets` so create and edit screens stay aligned. 
```
add_fieldsets = UserAdmin.add_fieldsets + (
    ("Assignments", {"fields": ("email", "role", "unit", "created_by")}),
)
```

mehalyna (commented):
The most critical risks are the broken DRF validators, the unsafe custom-user-model migration strategy, and emailing plaintext passwords. 
By principle: 
**SOLID** - serializer and service boundaries are blurred, permission logic depends on magic values; 
**SRP** - UserRegistrationSerializer.create does orchestration for user, profile, password, and email; 
**DRY** - role code literals and duplicated route mounting; 
**KISS** - auth setup is made more complex by replacing the user model and duplicating URL prefixes; 
**naming** - send_activation_email is misleading because it sends credentials rather than an activation flow; 
**readability** - multiline field wiring and mixed responsibilities make behavior harder to reason about.

yedynetsy (approved):
looks good overall, but there's a merge conflict that needs to be resolved before we can merge this

yedynetsy:
I see the pre-commit checks failed. Most of it was auto-fixed by Black and isort, but you need to manually remove a few unused imports in serializers.py, services.py, and views.py (check the Flake8 errors in the logs). After that, just run `pre-commit run --all-files` locally to make sure everything passes before pushing again

mehalyna (approved):

PR #44, opened 2026-05-09
Add Drone and DroneSpec models

torii044:
Added: 
- Drone and DroneSpec models 
- POST /api/drones/ endpoint
- tests for drone creation


drones/models.py
```diff
@@ -1,3 +1,39 @@
...
+    battery_capacity_mah = models.PositiveIntegerField()
+    camera_model = models.CharField(max_length=255)
+    vtx_model = models.CharField(max_length=255, blank=True)
+    flight_controller = models.CharField(max_length=255)
+    firmware_version = models.CharField(max_length=255, blank=True)
+    max_speed_kmh = models.DecimalField(max_digits=6, decimal_places=2)
+    max_range_km = models.DecimalField(max_digits=6, decimal_places=2)
+    max_flight_time = models.DecimalField(max_digits=6, decimal_places=2)
+    frequency_mhz = models.PositiveIntegerField()
+    payload_capacity_g = models.PositiveIntegerField(blank=True)
```
mehalyna: `payload_capacity_g` is misconfigured because `blank=True` alone does not make an integer field optional at the database/serializer level. Set `null=True` as well so omitted payload capacity values do not fail validation or migration constraints.

drones/migrations/0001_initial.py
```diff
@@ -0,0 +1,51 @@
...
+                ('battery_capacity_mah', models.PositiveIntegerField()),
+                ('camera_model', models.CharField(max_length=255)),
+                ('vtx_model', models.CharField(blank=True, max_length=255)),
+                ('flight_controller', models.CharField(max_length=255)),
+                ('firmware_version', models.CharField(blank=True, max_length=255)),
+                ('max_speed_kmh', models.DecimalField(decimal_places=2, max_digits=6)),
+                ('max_range_km', models.DecimalField(decimal_places=2, max_digits=6)),
+                ('max_flight_time', models.DecimalField(decimal_places=2, max_digits=6)),
+                ('frequency_mhz', models.PositiveIntegerField()),
+                ('payload_capacity_g', models.PositiveIntegerField(blank=True)),
```
mehalyna: The initial migration repeats the same optional-field bug and will create a non-null column for `payload_capacity_g`. Regenerate or edit the migration to keep schema and model behavior consistent.

drones/tests.py
```diff
@@ -1,3 +1,98 @@
...
+
+
+    def test_create_drone_duplicate_serial_number(self):
+        Drone.objects.create(
+            serial_number="Test Serial Number",
+            inventory_number="Test Inventory Number",
+            name="Drone",
+            status="ACTIVE",
+            acquired_at="2026-05-09",
+        )
```
mehalyna, line 77: This setup creates a Drone without the required model field, so the test can fail before it even exercises duplicate serial validation. Include all required model fields in the fixture to keep the test focused on the API behavior.

drones/serializers.py
```diff
@@ -0,0 +1,24 @@
...
+    class Meta:
+        model = Drone
+        fields = "__all__"
+
+    def create(self, validated_data):
+        spec_data = validated_data.pop("spec")
+        drone = Drone.objects.create(**validated_data)
+        DroneSpec.objects.create(drone=drone, **spec_data)
+
+        return drone
```
mehalyna: Nested object creation is not atomic, so a `Drone` can be persisted without a `DroneSpec` if the second insert fails. Wrap both writes in a transaction to avoid partial state and undue coupling between persistence steps. 
```
from django.db import transaction

def create(self, validated_data):
    with transaction.atomic():
        spec_data = validated_data.pop("spec")
        drone = Drone.objects.create(**validated_data)
        DroneSpec.objects.create(drone=drone, **spec_data)
        return drone
```

drones/admin.py
```diff
@@ -1,3 +1,15 @@
...
+class DroneAdmin(admin.ModelAdmin):
+    list_display = ("id", "serial_number", "inventory_number", "name", "model", "status")
+    search_fields = ("serial_number", "inventory_number", "name")
+    list_filter = ("status", "created_at")
+
+
+@admin.register(DroneSpec)
+class DroneSpecAdmin(admin.ModelAdmin):
+    list_display = ("id", "drone", "frame_type", "motor_model", "battery_type", "camera_model", "vtx_model")
+    search_fields = ("drone_id", "frame_type", "motor_model", "camera_model")
```
mehalyna: `search_fields = ("drone_id", ...)` is an anti-pattern because it searches the raw FK column instead of meaningful related drone attributes. Search through related fields so admins can find specs by serial or inventory number. 
```
@admin.register(DroneSpec) 
class DroneSpecAdmin(admin.ModelAdmin): 
    search_fields = ( "drone__serial_number", "drone__inventory_number", "frame_type", )
```

drones/tests.py
mehalyna, line 1: The test class covers only happy-path creation and a couple of failures, but it misses the optional nested fields that this serializer/model design depends on. Add a test omitting optional spec fields like `vtx_model`, `firmware_version`, and `payload_capacity_g` to catch the current validation/schema mismatch.

mehalyna (commented):
The most critical risks are a broken optional-field contract (payload_capacity_g), a non-atomic nested create that can leave partial data behind, and a test suite that currently masks at least one real defect.
Violations by principle: 
**SOLID** - serializer handles orchestration without transaction safety, increasing coupling to persistence details; 
**SRP** - the serializer’s create flow mixes validation result handling and multi-entity consistency concerns;
**DRY** - the same field bug is duplicated in both model and migration; 
**KISS** - admin search uses a low-level FK column instead of clear related attributes;
**naming** - model on Drone is ambiguous and easy to confuse with Django model terminology; 
**readability** - missing __str__ methods and weak tests reduce maintainability.

MykytaKa (approved):

drones/serializers.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework import serializers
```
yana-donda, line 1: I don't see the dependency and app registration added in this PR. Please add `djangorestframework` to `requirements.txt` and `rest_framework` to `INSTALLED_APPS` in config/settings.py

drones/views.py
```diff
@@ -1,3 +1,8 @@
-from django.shortcuts import render
+from rest_framework import generics
+from .models import Drone
+from .serializers import DroneSerializer
 
-# Create your views here.
+
+class DroneCreateView(generics.CreateAPIView):
+    queryset = Drone.objects.all()
+    serializer_class = DroneSerializer
```
yana-donda: Since drone registration is intended to be an administrator action, we should add explicit permissions here, for example `permission_classes = [IsAdminUser]` or a project-specific admin permission. Without this, the endpoint may be publicly accessible depending on global DRF settings.

drones/models.py
```diff
@@ -1,3 +1,45 @@
...
+    motor_model = models.CharField(max_length=255)
+    battery_type = models.CharField(max_length=255)
+    battery_capacity_mah = models.PositiveIntegerField()
+    camera_model = models.CharField(max_length=255)
+    vtx_model = models.CharField(max_length=255, blank=True)
+    flight_controller = models.CharField(max_length=255)
+    firmware_version = models.CharField(max_length=255, blank=True)
+    max_speed_kmh = models.DecimalField(max_digits=6, decimal_places=2)
+    max_range_km = models.DecimalField(max_digits=6, decimal_places=2)
+    max_flight_time = models.DecimalField(max_digits=6, decimal_places=2)
```
yana-donda: Could we include the unit in this field name, for example `max_flight_time_min` or `max_flight_time_minutes`? Other fields already include units like `_kmh`, `_km`, `_mhz`, and `_g`, so this would make the schema more consistent and avoid ambiguity.

yana-donda (approved):

mehalyna (approved):

PR #46, opened 2026-05-14
Add military_unit FK and System Admin permission

torii044:
- added a military_unit FK to the Drone model
- replaced IsAdminUser with Admin role permission


drones/tests.py
```diff
@@ -36,11 +37,9 @@ def setUp(self):
...
             },
         }
-        self.user = User.objects.create_user(
-            username="admin",
-            password="12345",
-            is_staff=True,
-        )
+        self.user = User.objects.create_user(username="admin", password="12345")
+
+        self.user.role = Role.objects.create(code=ADMIN_CODE, name="Admin")
```
mehalyna: 1. The test mutates self.user.role in memory but never persists it, so permission checks that reload the user from the database can still see no role. Save the relation after assignment.

```
self.user = User.objects.create_user(username="admin", password="12345")
self.user.role = Role.objects.create(code=ADMIN_CODE, name="Admin")
self.user.save(update_fields=["role"])

self.client.force_authenticate(self.user)
```

2. The test creates a role with ADMIN_CODE, which may not satisfy the new IsSystemAdmin permission and can make this test validate the wrong behavior. Use the exact role/permission code expected by IsSystemAdmin.

```
self.user = User.objects.create_user(username="admin", password="12345")
self.user.role = Role.objects.create(code=SYSTEM_ADMIN_CODE, name="System Admin")
self.user.save(update_fields=["role"])

self.client.force_authenticate(self.user)
```

3. The test now depends on concrete role-creation details, which is undue coupling to authorization internals and makes the setup harder to reuse. Move role/user creation into a helper or factory so permission fixtures stay consistent.

M1ro1 (approved):

drones/migrations/0002_drone_military_unit_alter_drone_id_and_more.py
```diff
@@ -0,0 +1,40 @@
...
+                to="accounts.militaryunit",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="drone",
+            name="id",
+            field=models.BigAutoField(
+                auto_created=True, primary_key=True, serialize=False, verbose_name="ID"
+            ),
+        ),
```
mehalyna: This migration includes unrelated id field alterations, which is a migration noise anti-pattern and couples this schema change to framework-generated churn. Regenerate or edit the migration to keep only the new foreign key operation.

drones/admin.py
```diff
@@ -11,10 +11,17 @@ class DroneAdmin(admin.ModelAdmin):
...
-    search_fields = ("serial_number", "inventory_number", "name")
-    list_filter = ("status", "created_at")
+    search_fields = (
+        "serial_number",
+        "inventory_number",
+        "name",
+        "military_unit__name",
+        "military_unit__code",
+    )
+    list_filter = ("status", "created_at", "military_unit__name")
```
mehalyna: Filtering on military_unit__name is brittle because admin list filters work best with the relation field, not a related text lookup. Filter by military_unit so Django uses the foreign key filter UI correctly.

drones/models.py
```diff
@@ -15,6 +17,13 @@ class Drone(models.Model):
     name = models.CharField(max_length=100)
     drone_model = models.CharField(max_length=255)
     status = models.CharField(max_length=20, choices=STATUS_CHOICES, default="ACTIVE")
+    military_unit = models.ForeignKey(
+        MilitaryUnit,
+        on_delete=models.PROTECT,
+        related_name="drones",
+        null=True,
+        blank=True,
+    )
```
mehalyna, line 22: Adding a nullable military_unit weakens the domain model and allows drones without ownership context, which can spread null-handling across the codebase.

drones/models.py
```diff
@@ -15,6 +17,13 @@ class Drone(models.Model):
     name = models.CharField(max_length=100)
     drone_model = models.CharField(max_length=255)
     status = models.CharField(max_length=20, choices=STATUS_CHOICES, default="ACTIVE")
+    military_unit = models.ForeignKey(
+        MilitaryUnit,
```
mehalyna: Importing MilitaryUnit directly increases coupling between apps and can create avoidable model import cycles. Reference the related model lazily by app label string instead.
```
    ...
    "accounts.MilitaryUnit",
    ...
```

yana-donda (approved):

mehalyna (approved):

PR #47, opened 2026-05-14
Add admin rbac role assignment

anastasiiahumeniuk:
This PR implements role-based access control foundations for user role management in the system.

It adds supported system roles, centralized RBAC permissions, secure role update logic, denied access logging, audit logging for role changes, and a role update endpoint.

## Changes

- added `Dispatcher` to supported roles
- added seed migration for default roles:
  - Admin
  - Commander
  - Operator
  - Technician
  - Viewer
  - Dispatcher
- introduced centralized RBAC permission matrix
- added reusable RBAC permission utilities and DRF permission classes
- added denied permission attempt logging
- implemented secure user role update service using `users.role_id`
- added validation for invalid or missing `role_id`
- prevented unsafe role changes:
  - self-demotion of admin users
  - changing the last admin to a non-admin role
  - removing admin role from the root/superuser account
- added audit log model for user role changes
- stored previous role and new role in audit log
- exposed role change audit log in Django admin
- improved Django admin role management for users and roles
- added `PATCH /accounts/users/{id}/role/` endpoint
- updated existing endpoints to use RBAC-based permission checks where applicable

### Update user role
`PATCH /accounts/users/{id}/role/`


accounts/services.py
```diff
@@ -57,3 +60,61 @@ def send_activation_email(user) -> None:
...
+
+    if locked_user.is_superuser and was_admin and not will_be_admin:
+        raise ValidationError(
+            {"role_id": ["Cannot remove the admin role from the root account."]}
+        )
+
+    if was_admin and not will_be_admin and count_admin_users() <= 1:
+        raise ValidationError(
+            {"role_id": ["Cannot change the last admin to a non-admin role."]}
+        )
```
mehalyna: The count_admin_users() check races with concurrent role updates and may still allow demoting the last admin. Requery admin count with select_for_update() inside the same transaction after locking the target user to guarantee consistency.

```
@transaction.atomic
def count_admin_users_locked() -> int:
    return User.objects.select_for_update().filter(role__code=ADMIN_CODE).count()

if was_admin and not will_be_admin and count_admin_users_locked() <= 1:
    raise ValidationError(...)
```

accounts/permissions.py
```diff
@@ -1,13 +1,124 @@
...
+        permission_code = self.required_permission or getattr(
+            view, "required_permission", None
+        )
+
+        if not permission_code:
+            log_permission_denied(
+                request,
+                permission_code,
+                reason="required_permission_not_configured",
+            )
```
mehalyna, line 84: Configuration errors (missing required_permission) are logged and silently denied as 403s, hiding developer mistakes in production. Raise ImproperlyConfigured to fail loudly during deployment instead of masking bugs.

accounts/permissions.py
```diff
@@ -1,13 +1,124 @@
...
+        permission_codes = self.required_permissions or getattr(
+            view, "required_permissions", None
+        )
+
+        if not permission_codes:
+            log_permission_denied(
+                request,
+                None,
+                reason="required_permissions_not_configured",
+            )
```
mehalyna, line 110: Same configuration-masking issue as HasRBACPermission - missing required_permissions is treated as a normal permission denial. Validate configuration explicitly to catch deployment errors early.

accounts/rbac.py
```diff
@@ -0,0 +1,146 @@
...
+        PERMISSION_MISSIONS_VIEW,
+        PERMISSION_MAINTENANCE_VIEW,
+        PERMISSION_WRITEOFF_VIEW,
+        PERMISSION_MEDIA_VIEW,
+        PERMISSION_AUDIT_LOGS_VIEW_OWN,
+        PERMISSION_PROFILE_VIEW_OWN,
+        PERMISSION_PROFILE_UPDATE_OWN,
+        PERMISSION_PROFILE_RESET_PASSWORD_OWN,
+    },
+}
```
mehalyna: This 150-line permission matrix creates a God object anti-pattern that tightly couples accounts to all role codes and makes permission audits difficult. Extract role-specific permission sets into composable functions or a dedicated policy provider.

```
def admin_permissions():
    return {PERMISSION_USERS_MANAGE_ROLES, PERMISSION_USERS_CREATE, ...}

ROLE_PERMISSION_MATRIX = {
    ADMIN_CODE: admin_permissions(),
    COMMANDER_CODE: commander_permissions(),
}
```

accounts/models.py
```diff
@@ -60,3 +60,43 @@ class UserProfile(models.Model):
...
+
+    class Meta:
+        ordering = ("-changed_at",)
+
+    def __str__(self) -> str:
+        previous_code = self.previous_role.code if self.previous_role else "None"
+        new_code = self.new_role.code if self.new_role else "None"
+        return (
+            f"Role change for {self.target_user_id}: " f"{previous_code} -> {new_code}"
+        )
```
mehalyna, line 108: The __str__ method only shows user and role IDs, making audit logs hard to interpret in admin panels and debugging sessions. Include human-readable usernames and role names for better operational visibility.

accounts/services.py
```diff
@@ -57,3 +60,61 @@ def send_activation_email(user) -> None:
...
+        raise ValidationError({"role_id": ["Invalid role_id."]}) from exc
+
+    locked_user = (
+        User.objects.select_for_update().select_related("role").get(pk=target_user.pk)
+    )
+    previous_role = locked_user.role
+
+    if previous_role_id := getattr(previous_role, "id", None):
+        if previous_role_id == new_role.id:
+            return locked_user
```
mehalyna, line 105: Audit logging is silently skipped when the role is unchanged, hiding attempted role changes that might indicate suspicious activity. Log all role-change requests including no-ops to maintain a complete audit trail.

```
if previous_role_id == new_role.id:
    UserRoleAuditLog.objects.create(
        changed_by=changed_by, target_user=locked_user,
        previous_role=previous_role, new_role=new_role,
    )
    return locked_user
```

roles/migrations/0003_seed_default_roles.py
```diff
@@ -0,0 +1,72 @@
...
+    Role.objects.filter(
+        code__in=[
+            "ADMIN",
+            "COMMANDER",
+            "OPERATOR",
+            "TECHNICIAN",
+            "VIEWER",
+            "DISPATCHER",
+        ]
+    ).delete()
```
mehalyna: The reverse migration deletes all matching roles unconditionally, which can destroy production data if users create custom roles with the same codes. Make the reverse operation a no-op to prevent data loss.

accounts/admin.py
```diff
@@ -20,25 +32,72 @@ class CustomUserAdmin(UserAdmin):
...
+    list_select_related = ("target_user", "previous_role", "new_role", "changed_by")
+
+    def has_add_permission(self, request):
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
```
mehalyna: Three identical permission methods with hard-coded False values violate DRY and reduce readability. Consolidate into a single helper method to simplify the admin class.

```
def _read_only_permission(self, request, obj=None):
    return False

has_add_permission = has_change_permission = has_delete_permission = _read_only_permission
```

accounts/views.py
```diff
@@ -1,19 +1,50 @@
...
+    permission_classes = [HasRBACPermission]
+    required_permission = PERMISSION_USERS_CREATE
+
+
+class UserRoleUpdateAPIView(APIView):
+    permission_classes = [HasRBACPermission]
+    required_permission = PERMISSION_USERS_MANAGE_ROLES
+
+    def patch(self, request, user_id):
+        target_user = get_object_or_404(User.objects.select_related("role"), pk=user_id)
```
mehalyna, line 32: The endpoint does not validate user active status or business rules before role updates, allowing role changes on deactivated accounts. Add pre-validation to reject updates on inappropriate users.

accounts/serializers.py
```diff
@@ -42,3 +42,25 @@ def create(self, validated_data):
         created_by = request.user if request and hasattr(request, "user") else None
 
         return create_user_account(validated_data, created_by)
+
+
+class UserRoleUpdateSerializer(serializers.Serializer):
+    role_id = serializers.IntegerField(required=True)
```
mehalyna: The serializer only validates role_id as an integer, deferring all business validation to the service layer, which makes the API contract unclear. Add basic validation (e.g., role_id > 0) to catch obvious errors early.

accounts/services.py
anastasiiahumeniuk: Fixed, thanks. I changed it to use a locked admin count inside the same transaction.

accounts/permissions.py
anastasiiahumeniuk: Fixed. Missing required_permissions now raises ImproperlyConfigured instead of being treated as a normal permission denial

accounts/rbac.py
anastasiiahumeniuk: I refactored the matrix into role-specific composable permission functions while keeping the centralized RBAC mapping.

accounts/views.py
anastasiiahumeniuk: Fixed: added pre validation to reject role updates for inactive users before applying the role change

M1ro1 (approved):

yedynetsy (approved):

Issue #23, opened 2026-05-05
As an operator, I can view assigned missions and update their status


mehalyna:
**User Story:**
As an operator, I want to see the list of missions that I am assigned to, view their details, and be able to update the mission status (e.g., In Progress, Completed, Aborted), so that mission progress is tracked and visible to all stakeholders.

**Technical tasks:**
- Define valid status transition matrix: planned → active or aborted, active → completed or aborted, completed and aborted are terminal — no further transitions allowed
- Implement GET /api/missions/?assigned_to=me endpoint with pagination — return missions where current user is in mission_drones.operator_id
- Implement PATCH /api/missions/{id}/status/ endpoint with transition matrix validation
- Reject invalid status transitions with a descriptive error response
- Set updated_at automatically on each status change
- On transition to active: update all assigned drones status to in_mission via drones table
- On transition to completed or aborted: update all assigned drones status back to active unless condition_after on mission_drones indicates otherwise — coordinate with drone condition recording task
- Add audit log entry on each status change via audit_logs table with action = mission_status_changed, target_model = Mission, changes containing previous and new status
- Add permission check: Operator can update only their assigned missions; Admin and Commander can update any

**Acceptance Criteria:**
- Operator can log in and view a dashboard/list of their assigned missions.
- Operator can see mission details, location, timing, and objectives.
- Operator can update the status of a mission with a valid status change.
- Status updates are timestamped and tracked in the mission history.

PR #48, opened 2026-05-15
Task#21 mission creation

olenakshyk:
Added a new missions app with model, REST endpoints, validation, permissions and admin support.

What was implemented:

- Mission model in missions/models.py
- Database migration for missions table
- POST /api/missions/ endpoint with serializer and validation for required fields: title, started_at, location_description or coordinates
- GET /api/missions/ endpoint with filtering by status and pagination
- GET /api/missions/{id}/ endpoint — full mission detail view
- Permission check: only Dispatcher or Admin can create missions


WWINMWWITW (approved):

missions/serializers.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def validate_commander_id(self, user):
+        if user is None:
+            return user
+        role_code = getattr(getattr(user, "role", None), "code", None)
+        if role_code != COMMANDER_CODE:
+            raise serializers.ValidationError(
+                "Selected user does not have the Commander role."
+            )
+        return user
```
mehalyna, line 88: Direct role code checking creates tight coupling to the roles app. Use Django permissions or a method on the User model instead.

missions/serializers.py
```diff
@@ -0,0 +1,87 @@
...
+            "created_by",
+            "created_at",
+            "updated_at",
+        ]
+
+    def validate_title(self, value):
+        value = (value or "").strip()
+        if not value:
+            raise serializers.ValidationError("Title is required.")
+        return value
```
mehalyna: Stripping whitespace is good, but the method modifies the value which might not persist. Return the stripped value explicitly and consider minimum length validation.

missions/models.py
```diff
@@ -0,0 +1,78 @@
...
+        blank=True,
+    )
+
+    location_description = models.TextField(null=True, blank=True)
+    latitude = models.DecimalField(
+        max_digits=9, decimal_places=6, null=True, blank=True
+    )
+    longitude = models.DecimalField(
+        max_digits=9, decimal_places=6, null=True, blank=True
+    )
```
mehalyna, line 59: Latitude max_digits=9 with decimal_places=6 allows values up to 999.999999, but valid latitude ranges from -90 to 90. 
https://www.gps-coordinates.net/#google_vignette
This will accept invalid coordinates.

missions/permissions.py
```diff
@@ -0,0 +1,20 @@
...
+            return False
+        if request.method in permissions.SAFE_METHODS:
+            return True
+        role = getattr(request.user, "role", None)
+        if role is not None and getattr(role, "code", None) in [
+            DISPATCHER_CODE,
+            ADMIN_CODE,
+        ]:
+            return True
+        return False
```
mehalyna: The permission logic tightly couples to role.code implementation details from another app. If roles.models changes, this breaks (Dependency Inversion Principle violation).

missions/serializers.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def validate_started_at(self, value):
+        if value is None:
+            raise serializers.ValidationError("started_at is required.")
+        if value < timezone.now():
+            raise serializers.ValidationError("started_at cannot be in the past.")
+        return value
```
mehalyna, line 79: The validation prevents scheduling missions in the past, but uses timezone.now() which can cause race conditions if creation takes time. Consider allowing a small grace period or move to model validation.

missions/serializers.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if not location and not has_coordinates:
+            raise serializers.ValidationError(
+                "Either location or latitude and longitude must be provided."
+            )
+
+        if (latitude is None) ^ (longitude is None):
+            raise serializers.ValidationError(
+                "latitude and longitude must be provided together."
+            )
```
mehalyna: The XOR check is clever but reduces readability. Use explicit boolean logic.

missions/models.py
```diff
@@ -0,0 +1,78 @@
...
+
+    started_at = models.DateTimeField()
```
mehalyna: started_at is required but can be set to any future datetime, while status defaults to "planned". No validation ensures status transitions are valid (planned -> active -> completed).

missions/views.py
```diff
@@ -0,0 +1,26 @@
...
+class MissionListCreateView(generics.ListCreateAPIView):
+    serializer_class = MissionSerializer
+    permission_classes = [permissions.IsAuthenticated, IsDispatcherOrAdmin]
+
+    def get_queryset(self):
+        queryset = Mission.objects.with_related()
+        status = self.request.query_params.get("status")
+        if status:
+            queryset = queryset.filter(status=status)
+        return queryset
```
mehalyna, line 27: Query parameter filtering on status lacks validation. Invalid status values will silently return empty results instead of returning an error.

missions/views.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def perform_create(self, serializer):
+        serializer.save(created_by=self.request.user)
+
+
+class MissionDetailView(generics.RetrieveAPIView):
+    serializer_class = MissionSerializer
+    permission_classes = [permissions.IsAuthenticated, IsDispatcherOrAdmin]
+    queryset = Mission.objects.with_related()
```
mehalyna, line 36: MissionDetailView only allows retrieve (GET), but the permission class IsDispatcherOrAdmin allows all safe methods. This is inconsistent - detail views should be accessible to all authenticated users if list view is.

missions/serializers.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class MissionSerializer(serializers.ModelSerializer):
+    commander_id = serializers.PrimaryKeyRelatedField(
+        source="commander",
+        queryset=User.objects.all(),
+        required=False,
+        allow_null=True,
+    )
+    created_by = serializers.PrimaryKeyRelatedField(read_only=True)
```
mehalyna: created_by is read_only but never gets serialized properly for nested representation. Users see only the ID instead of useful info like username or email.
```
class UserBriefSerializer(serializers.ModelSerializer):
    class Meta:
        model = User
        fields = ['id', 'username', 'email']

class MissionSerializer(serializers.ModelSerializer):
    created_by = UserBriefSerializer(read_only=True)
    commander = UserBriefSerializer(source='commander', read_only=True)
```

missions/models.py
olenakshyk: I think validation of started_at and status is part of task #23, so for now I'll make the started_at field optional

MaksymKorchynskyi (approved):

mehalyna (approved):

PR #49, opened 2026-05-16
Task#37 implement user activation/deactivation logic and audit log

yedynetsy:
**What was done:**

- Added UserStatusUpdateView to toggle user is_active status (Admin only).

- Implemented Session invalidation: active sessions are deleted when a user is deactivated.

- Created UserStatusLog model to track who changed user statuses and why.

- Updated Django Admin (accounts/admin.py) to include an is_active toggle and display the audit logs.


accounts/models.py
```diff
@@ -62,6 +62,25 @@ def __str__(self) -> str:
...
+        null=True,
+        related_name="performed_status_changes",
+    )
+    old_status = models.BooleanField()
+    new_status = models.BooleanField()
+    reason = models.TextField(blank=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self):
+        return f"Change for {self.target_user} by {self.changed_by}"
```
anastasiiahumeniuk: The current string representation is valid, but it could be more descriptive for debugging and admin display.
Consider including the performed action and timestamp, for example:
"User deactivated by admin1 on 2026-05-16"
This would make audit log entries easier to read without opening the full object.

M1ro1 (approved):

anastasiiahumeniuk (approved):

mehalyna (approved):

PR #50, opened 2026-05-17
Add collecting logs for different user's actions

M1ro1:
## Description
This PR introduces a comprehensive Audit Logging system to track user actions, logins, and system events, ensuring compliance and security monitoring as per the user story. 

## Technical Changes & Tasks Completed
- **Data Layer:** - Created the `AuditLog` model with fields for `actor`, `target_user`, `action_type`, `result`, `ip_address`, and `user_agent`.
  - Enforced immutability at the model level (overrode `save` and `delete` methods).
- **Admin Panel:** Registered `AuditLog` in Django Admin as strictly read-only.
- **Logging Integration:**
  - Implemented `create_audit_log` service helper.
  - Added Django Signals to automatically track `LOGIN_SUCCESS`, `LOGIN_FAILED`, and `LOGOUT`.
  - Integrated explicit logging into `create_user_account` and `update_user_role` services.
- **API (DRF):**
  - Created `AuditLogViewSet` using `ReadOnlyModelViewSet` to prevent unauthorized updates/deletions endpoints.
  - Implemented Row-Level Security: Admins can view all logs; users can only view logs where they are the actor or target.
  - Added `django-filter` support for filtering by `actor`, `target_user`, `action_type`, `result`, and `created_at`.
  - Added standard pagination.
  - Implemented a custom `/accounts/audit-logs/export/` endpoint to download filtered logs as a CSV file.

## How to Test
1. **Migrations:** Run `python manage.py makemigrations` and `python manage.py migrate`.
2. **Auth Signals:** Log in and out (including a failed login attempt with a wrong password). Check the database or Admin panel to see the generated logs.
3. **Services:** Create a new user and change a user's role via API. Verify that `USER_CREATED` and `ROLE_CHANGED` events are logged.
4. **Permissions:** - Fetch `/accounts/audit-logs/` with an Admin  -> Should return all logs.
   - Fetch `/accounts/audit-logs/` with a standard User  -> Should return only logs related to that user.
5. **Filters & Export:** Test query parameters (e.g., `?action_type=LOGIN_SUCCESS`) and hit the `/accounts/audit-logs/export/` endpoint to verify the CSV file downloads correctly.


anastasiiahumeniuk (approved):

yedynetsy (approved):

accounts/models.py
```diff
@@ -106,3 +106,72 @@ def __str__(self) -> str:
...
+        actor_name = self.actor.username if self.actor else "System/Anonymous"
+        return (
+            f"[{self.created_at.strftime('%Y-%m-%d %H:%M:%S')}] "
+            f"{actor_name} - {self.action_type} ({self.result})"
+        )
+
+    def save(self, *args, **kwargs):
+        if self.pk is not None:
+            raise ValueError("AuditLog entries are immutable and cannot be updated.")
+        super().save(*args, **kwargs)
```
mehalyna, line 213: The immutability check allows creation but prevents updates, however this check happens after validation which wastes resources. Move the check before calling super().save() and allow updates during creation by checking if it's a real update attempt.

accounts/models.py
```diff
@@ -106,3 +106,72 @@ def __str__(self) -> str:
...
+
+    def delete(self, *args, **kwargs):
+        raise ValueError("AuditLog entries are immutable and cannot be deleted.")
```
mehalyna, line 216: The delete method raises an error but doesn't follow Django's expected signature which should return the deletion count. Override at the QuerySet level instead for proper bulk delete protection.

accounts/services.py
```diff
@@ -132,4 +140,39 @@ def update_user_role(*, target_user: User, new_role_id: int, changed_by: User) -
...
+    user_agent = ""
+
+    if request:
+        user_agent = request.META.get("HTTP_USER_AGENT", "")
+
+        x_forwarded_for = request.META.get("HTTP_X_FORWARDED_FOR")
+        if x_forwarded_for:
+            ip_address = x_forwarded_for.split(",")[0].strip()
+        else:
+            ip_address = request.META.get("REMOTE_ADDR")
```
mehalyna, line 168: IP address extraction from X-Forwarded-For header is vulnerable to spoofing since it trusts the first value without validation. Malicious users can inject arbitrary IPs. Take the last trusted proxy IP, not the first

accounts/signals.py
```diff
@@ -0,0 +1,45 @@
...
+        description="User logged in successfully",
+        request=request,
+    )
+
+
+@receiver(user_login_failed)
+def log_user_login_failed(sender, credentials, request, **kwargs):
+    username = credentials.get("username", "Unknown")
+    create_audit_log(
+        actor=None,
```
mehalyna, line 39: Setting actor=None for failed login attempts loses valuable information. If a user exists in the database with that username, reference them to track repeated failed attempts per account.

accounts/views.py
```diff
@@ -73,3 +80,72 @@ def post(self, request, user_id, token):
...
+    filterset_class = AuditLogFilter
+
+    def get_queryset(self):
+        user = self.request.user
+        if user.is_staff or user.is_superuser:
+            return AuditLog.objects.all().select_related("actor", "target_user")
+
+        return AuditLog.objects.filter(actor=user) | AuditLog.objects.filter(
+            target_user=user
+        ).select_related("actor", "target_user")
```
mehalyna: Using OR operator creates a union, potentially returning duplicate records if a user is both actor and target. Use distinct() or Q objects properly.

accounts/views.py
```diff
@@ -73,3 +80,72 @@ def post(self, request, user_id, token):
...
+
+    @action(detail=False, methods=["get"])
+    def export(self, request):
+        queryset = self.filter_queryset(self.get_queryset())
```
mehalyna: The export endpoint has no rate limiting or query size limits, allowing users to export potentially millions of records and cause memory/performance issues.

accounts/admin.py
```diff
@@ -99,3 +99,32 @@ def _read_only_permission(self, request, obj=None):
...
+        "ip_address",
+    )
+    list_filter = ("action_type", "result", "created_at", "ip_address")
+    search_fields = (
+        "actor__username",
+        "target_user__username",
+        "description",
+        "ip_address",
+    )
+    readonly_fields = [f.name for f in AuditLog._meta.fields]
```
mehalyna, line 154: Making all fields read-only is good, but the admin interface still allows access to the add button in the UI even though has_add_permission returns False. Add def get_actions to prevent bulk operations.
```
def get_actions(self, request):
    actions = super().get_actions(request)
    if 'delete_selected' in actions:
        del actions['delete_selected']
    return actions
```

mehalyna (approved):

PR #53, opened 2026-05-21
Implements mission assignment workflow for dispatchers

MaksymKorchynskyi:
- `MissionDrone` model with string-based FK to `drones.Drone` and User FK for operator
- Migration for `mission_drones` table (depends on `drones.0002`)
- `GET /api/missions/{id}/assignments/` — list assignments
- `POST /api/missions/{id}/assignments/` — create assignment
- `DELETE /api/missions/{id}/assignments/{id}/` — delete (only if mission is `planned`)
- Drone availability check (`status = ACTIVE`)
- Operator time-overlap check against active/planned missions
- On deletion: drone status reverted to `ACTIVE`
- Permission: Dispatcher and Admin only
- Audit logging via `audit_logs` table (`assignment_created` / `assignment_deleted`)


missions/serializers.py
```diff
@@ -106,3 +110,133 @@ def validate(self, attrs):
...
+
+        if not mission:
+            raise serializers.ValidationError(
+                {"mission": "Mission context is required for assignment validation."}
+            )
+
+        if mission.status != Status.PLANNED:
+            raise serializers.ValidationError(
+                {"mission": "Assignments can only be added to planned missions."}
+            )
```
olenakshyk: MissionDroneSerializer.validate() checks mission.status != PLANNED on the in-memory object passed via context, and create() never re-fetches it with select_for_update() inside the transaction. Only Drone and User (operator) are locked — Mission is not.

This leaves a TOCTOU window: between validate() and the INSERT, another request can transition the mission planned → active, and we'll still create an assignment on a non-planned mission, violating the invariant enforced everywhere else.

missions/serializers.py
```diff
@@ -106,3 +110,133 @@ def validate(self, attrs):
...
+        if operator:
+            overlapping = overlapping.filter(operator=operator)
+        if drone:
+            overlapping = overlapping.filter(drone=drone)
+
+        m_start = mission.started_at
+        m_end = mission.ended_at
+
+        if not m_start:
+            return overlapping.exists()
```
olenakshyk: It’s a bit harsh to assume that a mission without a start time will inevitably overlap with another one. I think it would be better to prohibit assigning missions without a start time to drones and operators and to return a clear error message about this.

missions/serializers.py
```diff
@@ -106,3 +110,133 @@ def validate(self, attrs):
...
+
+        if drone and mission:
+            if self._check_overlap(mission=mission, drone=drone):
+                raise serializers.ValidationError(
+                    {"drone": "Drone is assigned to another mission during this time."}
+                )
+
+        return attrs
+
+    def create(self, validated_data):
```
olenakshyk, line 152: MissionDroneSerializer.create() is doing far more than serialization — it locks rows, runs overlap checks, mutates drone.status, and writes audit logs. Meanwhile, the symmetric delete logic lives in MissionAssignmentDetailView.perform_destroy(). The two halves of the same domain operation now sit in different layers, which makes the code harder to test, harder to reuse, and easy to drift.

I'd suggest extracting this into a missions/services.py module with two functions: assign_drone_to_mission(...) and unassign_drone_from_mission(...). Both views call into the service; the serializer keeps only declarative validation (field types, role checks, required fields). _check_overlap should also move to the service module — it's a domain rule about resource scheduling, not a serializer concern, and we'll need it elsewhere (e.g. an "available drones" endpoint).

missions/serializers.py
```diff
@@ -106,3 +110,133 @@ def validate(self, attrs):
...
+            "mission",
+            "drone",
+            "operator",
+            "condition_after",
+            "condition_description",
+            "flight_started_at",
+            "flight_ended_at",
+            "created_at",
+        ]
+        read_only_fields = ["id", "mission", "created_at"]
```
olenakshyk, line 139: MissionDroneSerializer returns drone and operator as integers. UserBriefSerializer already has a convenient short format—we should add read-only nested fields (drone_details, operator_details) to facilitate GET requests; otherwise, the frontend will have to make N+1 requests.

olenakshyk (changes requested):

missions/services.py
```diff
@@ -0,0 +1,153 @@
...
+    m_end = mission.ended_at
+
+    q_objects = Q()
+    if m_end:
+        q_objects &= Q(mission__started_at__lt=m_end)
+    q_objects &= Q(mission__ended_at__isnull=True) | Q(
+        mission__ended_at__gt=m_start,
+    )
+
+    return overlapping.filter(q_objects).exists()
```
mehalyna: The overlap check logic has a bug when both missions have an ended_at value, allowing invalid concurrent assignments. The condition should ensure proper interval overlap by checking that the new mission starts before the existing mission ends AND the new mission ends after the existing mission starts.

missions/services.py
```diff
@@ -0,0 +1,153 @@
...
+            action="assignment_created",
+            target_model="MissionDrone",
+            user=action_user,
+            changes={
+                "mission_id": instance.mission_id,
+                "drone_id": instance.drone_id,
+                "operator_id": instance.operator_id,
+            },
+        )
+        return instance
```
mehalyna, line 126: The function performs redundant validation checks that were already done in the serializer, adding unnecessary database queries. Remove the duplicate validation logic and rely on the serializer's validation before calling this service function.

missions/serializers.py
mehalyna: Totally agree!

missions/models.py
```diff
@@ -85,3 +85,61 @@ class Meta:
...
+
+    class Meta:
+        db_table = "mission_drones"
+        constraints = [
+            models.UniqueConstraint(
+                fields=["mission", "drone"], name="unique_mission_drone"
+            ),
+            models.UniqueConstraint(
+                fields=["mission", "operator"], name="unique_mission_operator"
+            ),
```
mehalyna: The model has two conflicting unique constraints that make it impossible to assign multiple drones to one mission with different operators. The unique_mission_operator constraint should be removed to allow proper many-to-many assignment flexibility.

olenakshyk (approved):

WWINMWWITW (approved):

mehalyna (approved):

PR #55, opened 2026-05-22
add drone pagination, search, filtering 

torii044:
Added:
- classification field in Drone model 
- pagination, search, filtering, ordering for drones


drones/filters.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class DroneFilter(django_filters.FilterSet):
+    class Meta:
+        model = Drone
+        fields = {
+            "serial_number": ["icontains"],
+            "inventory_number": ["icontains"],
+            "status": ["exact"],
+            "drone_model": ["icontains"],
```
mehalyna: Using icontains lookup on drone_model field allows SQL injection-like behavior and causes performance issues without database indexing. Use exact lookup or add a separate search functionality with proper text search indexes.

drones/views.py
```diff
@@ -1,13 +1,25 @@
...
     serializer_class = DroneSerializer
     permission_classes = [DronePermission]
+    filter_backends = (
+        DjangoFilterBackend,
+        filters.SearchFilter,
+        filters.OrderingFilter,
+    )
+    filterset_class = DroneFilter
+    pagination_class = StandardResultsSetPagination
+    search_fields = ["serial_number", "inventory_number"]
```
mehalyna: The search_fields and filter fields overlap, causing redundant query parameters and user confusion. Remove serial_number and inventory_number from search_fields since they're already available as filters with icontains.

drones/models.py
```diff
@@ -34,10 +34,19 @@ class Drone(models.Model):
...
+        ("CLASS_3", "Class 3"),
+    ]
+
     serial_number = models.CharField(max_length=100, unique=True)
     inventory_number = models.CharField(max_length=100, unique=True)
     name = models.CharField(max_length=100)
     drone_model = models.CharField(max_length=255)
+    classification = models.CharField(
+        max_length=20, choices=CLASSIFICATION_CHOICES, default="CLASS_1"
+    )
```
mehalyna, line 49: The classification field has a hardcoded default value without business justification, potentially causing incorrect data categorization. Make the field required without a default to force explicit classification during drone creation.

common/pagination.py
mehalyna, line 1: The pagination class is defined in the drones app but is generic enough to be reused across the project. Move this to a common/shared app or config directory to avoid duplication when other apps need pagination.

drones/tests.py
```diff
@@ -303,3 +304,94 @@ def test_existing_writeoff_record_is_not_overwritten(self):
...
+            self.assertEqual(item["status"], "LOST")
+
+    def test_filtering_icontains_field(self):
+        response = self.client.get(
+            self.create_url, {"drone_model__icontains": "model_4"}
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        for item in response.data["results"]:
+            self.assertIn("model_4", item["drone_model"])
```
mehalyna: The test uses a hardcoded filter field name that doesn't match the FilterSet configuration syntax. Use the correct django-filters syntax with double underscore for lookup types.

drones/migrations/0005_drone_classification.py
```diff
@@ -0,0 +1,26 @@
...
+            name="classification",
+            field=models.CharField(
+                choices=[
+                    ("CLASS_1", "Class 1"),
+                    ("CLASS_2", "Class 2"),
+                    ("CLASS_3", "Class 3"),
+                ],
+                default="CLASS_1",
+                max_length=20,
+            ),
```
mehalyna, line 24: Adding a non-nullable field with a default to an existing table can cause performance issues on large datasets. Consider making it nullable first, backfilling data, then adding the constraint in a separate migration.

drones/views.py
```diff
@@ -1,13 +1,25 @@
...
+    ordering_fields = ["created_at", "status", "name"]
```
mehalyna: The ordering_fields list is missing classification field which was just added and would be useful for sorting drones by class. Add it to allow users to order by classification.

drones/serializers.py
```diff
@@ -121,6 +121,7 @@ class Meta:
             "inventory_number",
             "name",
             "drone_model",
+            "classification",
```
mehalyna, line 124: The classification field is added to the serializer fields list but there's no validation to ensure the classification matches the drone_model capabilities. Add custom validation to enforce business rules between these fields.

drones/filters.py
```diff
@@ -0,0 +1,16 @@
...
+            "classification": ["exact"],
+            "military_unit": ["exact"],
```
mehalyna, line 15: The military_unit filter uses exact lookup requiring users to know the unit ID, making it difficult to use. Add icontains lookup or a separate name-based filter for better UX.

drones/tests.py
```diff
@@ -303,3 +304,94 @@ def test_existing_writeoff_record_is_not_overwritten(self):
...
+        for i in range(self.page_size + 1):
+            if i % 2 == 0:
+                self.drones.append(DroneFactory(military_unit=self.military_unit))
+            else:
+                self.drones.append(
+                    DroneFactory(military_unit=self.military_unit, status="LOST")
+                )
+
+        self.user = ViewerUserFactory()
+        self.client.force_authenticate(self.user)
```
mehalyna, line 318: Creating page_size + 1 drones in a loop without bulk_create is inefficient and slows down test execution. Use DroneFactory.create_batch() or bulk_create for better performance.

drones/views.py
```diff
@@ -1,13 +1,25 @@
...
+from .filters import DroneFilter
 from .models import Drone
+from .pagination import StandardResultsSetPagination
 from .permissions import DronePermission
 from .serializers import DroneSerializer, DroneUpdateSerializer
 
 
-class DroneCreateView(generics.ListCreateAPIView):
+class DroneListCreateView(generics.ListCreateAPIView):
     serializer_class = DroneSerializer
```
yana-donda, line 13: For the list endpoint we probably should not use the full DroneSerializer, because it includes nested spec, writeoff_record and status_history. This can make the drones list response unnecessarily heavy, especially when drones have many status history records. It would be better to add a lightweight DroneListSerializer for GET list responses and keep the full serializer for retrieve/create/update.

drones/filters.py
```diff
@@ -0,0 +1,16 @@
...
+from .models import Drone
+
+
+class DroneFilter(django_filters.FilterSet):
+    class Meta:
+        model = Drone
+        fields = {
+            "serial_number": ["icontains"],
+            "inventory_number": ["icontains"],
+            "status": ["exact"],
```
yana-donda, line 12: There is a potential conflict between the status filter and the base queryset. The filter allows exact status filtering, but get_queryset (line 24 in drones/views.py) always excludes Drone.INACTIVE_STATUSES. As a result, requests like ?status=WRITTEN_OFF or ?status=SOLD will always return an empty list even if such drones exist. I think we need a way to include inactive statuses when requested.

drones/tests.py
yana-donda, line 1: Since this PR adds classification to the Drone model and exposes it through filtering/API responses, it would be good to add explicit tests for it. For example: create drones with different classifications, request ?classification=CLASS_1, and assert that only drones with CLASS_1 are returned. It would also be useful to assert that classification is present in the response.

mehalyna (approved):

yana-donda (approved):

MaksymKorchynskyi (approved):

PR #56, opened 2026-05-22
add mission status transitions, operator endpoints, and audit logs

WWINMWWITW:
- Enforced valid mission status transition matrix (with block on terminal states).
- Implemented GET /api/missions/?assigned_to=me endpoint with pagination for operators.
- Implemented PATCH /api/missions/{id}/status/ endpoint with strict transition validation.
- Added role-based permission checks (Operators can only update assigned missions; Admins and Commanders can update any).
- Integrated automated, cascading state updates to synchronize drone availability with mission lifecycle phases (including condition_after overrides).
- Added detailed audit log entries tracking previous and new states on each mission status change.


missions/views.py
```diff
@@ -34,3 +50,44 @@ class MissionDetailView(generics.RetrieveAPIView):
...
+
+            for link in mission_drones:
+                drone = link.drone
+
+                if link.condition_after:
+                    drone.status = link.condition_after
+                else:
+                    drone.status = "ACTIVE"
+
+                drone.save()
```
mehalyna: The drone status update logic creates N+1 queries and lacks transaction safety, risking partial updates if an error occurs. Wrap the entire operation in a transaction and use bulk_update for efficiency.

missions/views.py
```diff
@@ -34,3 +50,44 @@ class MissionDetailView(generics.RetrieveAPIView):
...
+            if assigned_drones_ids:
+                Drone.objects.filter(id__in=assigned_drones_ids).update(
+                    status="IN_MISSION"
+                )
+
+        elif mission.status in [Status.COMPLETED, Status.ABORTED]:
+            mission_drones = mission.mission_drones.all()
+
+            for link in mission_drones:
+                drone = link.drone
```
mehalyna, line 107: The loop-based drone status update creates N database queries instead of one bulk operation. Use bulk_update to update all drones in a single query for better performance.

```
elif mission.status in [Status.COMPLETED, Status.ABORTED]:
    mission_drones = mission.mission_drones.select_related('drone')
    drones_to_update = [link.drone for link in mission_drones]
    for link in mission_drones:
        link.drone.status = link.condition_after or "ACTIVE"
    Drone.objects.bulk_update(drones_to_update, ['status'])
```

missions/models.py
```diff
@@ -85,3 +96,49 @@ class Meta:
...
+
+
+class MissionDrone(models.Model):
+    mission = models.ForeignKey(
+        Mission, on_delete=models.CASCADE, related_name="mission_drones"
+    )
+
+    drone = models.ForeignKey(
+        "drones.Drone", on_delete=models.CASCADE, related_name="mission_participations"
+    )
```
mehalyna, line 112: Using CASCADE delete on the drone foreign key means deleting a drone will delete mission history, losing critical audit data. Change to PROTECT to prevent accidental deletion of drones with mission history.

missions/serializers.py
```diff
@@ -23,7 +24,37 @@ class Meta:
...
+        if user is None:
+            return user
+
+        role_code = getattr(getattr(user, "role", None), "code", None)
+
+        if role_code != OPERATOR_CODE:
+            raise serializers.ValidationError(
+                "Selected user does not have the Operator role."
+            )
+        return user
```
mehalyna, line 61: The operator role validation uses fragile nested getattr calls that return None silently, potentially allowing invalid users. Raise a clear validation error if role is missing.

missions/permissions.py
```diff
@@ -13,3 +13,25 @@ def has_permission(self, request, view):
...
+        ):
+            return False
+
+        role_name = request.user.role.name.upper()
+
+        if role_name in ["ADMIN", "COMMANDER"]:
+            return True
+
+        if role_name == "OPERATOR":
+            is_assigned = obj.mission_drones.filter(operator=request.user).exists()
```
mehalyna: Hardcoding role names as strings ('ADMIN', 'COMMANDER') creates tight coupling and will break if role names change. Use role codes from the roles.models constants instead.

missions/serializers.py
```diff
@@ -61,6 +93,20 @@ class Meta:
...
+        mission = Mission.objects.create(**validated_data)
+
+        for drone_item in drones_data:
+            MissionDrone.objects.create(
+                mission=mission,
+                drone=drone_item["drone"],
+                operator=drone_item.get("operator"),
+            )
+
+        return mission
```
mehalyna, line 121: The mission creation with drones has no validation for drone availability or operator conflicts, allowing double-booking. Add validation to check if drones/operators are already assigned to overlapping missions.

missions/models.py
```diff
@@ -85,3 +96,49 @@ class Meta:
...
+class MissionAuditLog(models.Model):
+    user = models.ForeignKey(
+        "accounts.User",
+        on_delete=models.SET_NULL,
+        null=True,
+    )
+    action = models.CharField(max_length=255)
+    target_model = models.CharField(max_length=100)
+    target_id = models.IntegerField()
+    changes = models.JSONField(default=dict)
```
mehalyna: Using JSONField with default=dict is dangerous because mutable defaults are shared across instances, potentially causing data leakage. Use a callable default factory instead.

missions/views.py
```diff
@@ -24,6 +34,12 @@ def get_queryset(self):
                     }
                 )
             queryset = queryset.filter(status=status)
+
+        assigned_to = self.request.query_params.get("assigned_to")
+        if assigned_to == "me":
+            user = self.request.user
+            queryset = queryset.filter(mission_drones__operator_id=user.id).distinct()
```
mehalyna: The assigned_to=me filter lacks validation and only checks for string equality, allowing potential injection. Validate the parameter value explicitly before using it.

missions/admin.py
```diff
@@ -1,6 +1,12 @@
 from django.contrib import admin
 
-from .models import Mission
+from .models import Mission, MissionAuditLog, MissionDrone
+
+
+class MissionDroneInline(admin.TabularInline):
+    model = MissionDrone
+    extra = 0
+    fields = ("drone", "condition_after")
```
mehalyna: The inline only shows drone and condition_after fields, omitting the critical operator field from the admin interface. Add operator to the fields tuple for complete mission assignment management.

missions/serializers.py
```diff
@@ -106,3 +152,19 @@ def validate(self, attrs):
...
+class MissionStatusUpdateSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Mission
+        fields = ["status"]
+
+    def validate_status(self, value):
+        current_status = self.instance.status
+        allowed_transions = MISSION_STATUS_TRANSITIONS.get(current_status, [])
+
+        if value not in allowed_transions:
```
mehalyna: The typo in variable name allowed_transions (should be transitions) will cause confusion and makes the code less maintainable. Fix the spelling to match standard conventions.

missions/permissions.py
```diff
@@ -13,3 +13,25 @@ def has_permission(self, request, view):
...
+            not request.user
+            or not request.user.is_authenticated
+            or not request.user.role
+        ):
+            return False
+
+        role_name = request.user.role.name.upper()
+
+        if role_name in [ADMIN_CODE, COMMANDER_CODE]:
+            return True
```
olenakshyk: Permission check uses the wrong role identifier.  Role.name is a free-form display field ("Admin", "System Administrator", etc.); the canonical identifier is role.code (and the rest of the codebase uses get_user_role_code()). This breaks the moment a role's name diverges from its code, silently denying or granting access.

missions/views.py
```diff
@@ -34,3 +55,49 @@ class MissionDetailView(generics.RetrieveAPIView):
...
+
+class MissionStatusUpdateView(generics.RetrieveUpdateAPIView):
+    serializer_class = MissionStatusUpdateSerializer
+    permission_classes = [permissions.IsAuthenticated, CanUpdateMissionStatus]
+    queryset = Mission.objects.all()
+
+    def perform_update(self, serializer):
+        with transaction.atomic():
+            old_status = serializer.instance.status
+            mission = serializer.save()
```
olenakshyk: perform_update reads old_status, validates the transition, then saves — without select_for_update(). Two concurrent PATCH /status/ calls on the same mission can both pass the matrix check (e.g. both planned → active), write the audit log twice, and double-update drone statuses.

missions/serializers.py
```diff
@@ -105,4 +157,63 @@ def validate(self, attrs):
...
+
+        if drone_ids or operator_ids:
+            time_overlap = Q(mission__started_at__lte=ended_at) if ended_at else Q()
+            time_overlap &= Q(mission__ended_at__gte=started_at) | Q(
+                mission__ended_at__isnull=True
+            )
+
+            conflicting_links = MissionDrone.objects.filter(
+                mission__status__in=[Status.PLANNED, Status.ACTIVE]
+            ).filter(time_overlap)
```
olenakshyk, line 205: The drone/operator booking check queries MissionDrone across PLANNED/ACTIVE missions without excluding self.instance on updates. Editing an existing mission's other fields (where its drones come back through attrs) will report the mission's own drones/operators as already booked, blocking legitimate edits.

missions/serializers.py
```diff
@@ -105,4 +157,63 @@ def validate(self, attrs):
...
+        drone_ids = [item["drone"].id for item in drones_data if "drone" in item]
+        operator_ids = [
+            item["operator"].id for item in drones_data if item.get("operator")
+        ]
+
+        if drone_ids or operator_ids:
+            time_overlap = Q(mission__started_at__lte=ended_at) if ended_at else Q()
+            time_overlap &= Q(mission__ended_at__gte=started_at) | Q(
+                mission__ended_at__isnull=True
+            )
```
olenakshyk, line 201: The check doesn't handle cases where started_at or ended_at is None correctly. I think it would be better to use the function for checking overlaps from PR task#22 once it's merged.

missions/serializers.py
```diff
@@ -61,6 +99,20 @@ class Meta:
             "updated_at",
         ]
 
+    def create(self, validated_data):
```
olenakshyk, line 109: MissionSerializer defines create() to persist mission_drones, but no update(). A PATCH/PUT carrying drones=[…] will pass validation (including the self-conflict bug), silently discard the payload, and return a stale response - data-loss class bug for the dispatcher edit flow.

olenakshyk (approved):

MaksymKorchynskyi (approved):

mehalyna (approved):

Issue #24, opened 2026-05-05
As an operator, I can record mission outcome and link results to drone condition


mehalyna:
**User Story:**
As an operator, upon completing a mission, I want to record the mission outcome (success/failure, observations, incident notes), and link mission results to the health/condition of the drone, so that post-mission assessments are comprehensive and actionable.

**Technical tasks:**
- Implement PATCH /api/missions/{id}/outcome/ endpoint — record result and notes on existing Mission record
- Restrict outcome recording to missions with status completed or aborted
- Implement PATCH /api/missions/{id}/assignments/{assignment_id}/condition/ endpoint — record condition_after and condition_description on existing MissionDrone record
- Define allowed values for condition_after: ok, damaged, lost
- On condition_after = ok: update drones.status back to active
- On condition_after = damaged: update drones.status to damaged
- On condition_after = lost: update drones.status to written_off — coordinate with writeoff team on whether a writeoff_records entry should be created automatically or manually
- Create DroneStatusHistory record on each drone status change with related_mission_id — coordinate with drones team on who owns this model; if already created in drones app, use it directly
- Add permission check: only assigned Operator or Admin can record outcome
- Add audit log entry on outcome recording via audit_logs table with action = mission_outcome_recorded, target_model = 
- Mission, changes containing result and previous and new drone condition

**Acceptance Criteria:**
- Operator can input mission outcome details after completion.
- Operator can record drone condition post-mission (checklist or note).
- The system updates drone records with relevant data from the mission.
- All mission reports and associated drone health notes are accessible for future reference.

PR #61, opened 2026-05-28
Task#24 mission outcome

olenakshyk:
## Summary

Implements task #24 — mission outcome recording and post-mission drone condition tracking.

### Endpoints
- `PATCH /api/missions/{id}/outcome/` — record `result`, `notes`, `incident_notes` on a completed/aborted mission.
- `PATCH /api/missions/{id}/assignments/{assignment_id}/condition/` — record `condition_after` and `condition_description` on a `MissionDrone`, and propagate the change to the drone.

### What was implemented
- New permission `IsAssignedOperatorOrAdmin` — only the assigned operator (or Admin) can record outcome / condition.
- Service `record_mission_outcome()` — locks the mission, validates status, writes `AuditLog` (`mission_outcome_recorded`).
- Service `record_drone_condition()` — maps condition → drone status (`ok→ACTIVE`, `damaged→DAMAGED`, `lost→WRITTEN_OFF`) and delegates to `drones.services.update_drone`, which automatically creates `DroneStatusHistory` and (for `lost`) a `WriteOffRecord` linked to the mission.
- Test suite (`missions/tests.py`) covering success/forbidden/validation paths for both endpoints, plus drone status, history, and write-off side effects


missions/views.py
MaksymKorchynskyi: There's a conflict between this block and the new condition flow. condition_after now stores Condition values ("ok", "damaged", "lost"), but here it's assigned directly to drone.status, which expects Drone.STATUS_CHOICES ("ACTIVE", "DAMAGED", "WRITTEN_OFF"). If a condition is recorded before the status transition, the drone ends up with an invalid status. bulk_update also bypasses update_drone(), so no DroneStatusHistory or WriteOffRecord is created. Worth either routing through CONDITION_TO_DRONE_STATUS + update_drone(), or dropping this logic entirely in favour of the new /condition/ endpoint.

missions/views.py
```diff
@@ -62,6 +86,33 @@ class MissionDetailView(generics.RetrieveAPIView):
...
+        MissionsRecordConditionRBAC,
+        IsAssignedOperatorOrAdmin,
+    ]
+    lookup_url_kwarg = "assignment_id"
+    http_method_names = ["patch", "options", "head"]
+
+    def get_queryset(self):
+        return MissionDrone.objects.filter(
+            mission_id=self.kwargs["pk"],
+        ).select_related("mission", "drone", "operator")
```
MaksymKorchynskyi, line 119: For MissionDroneConditionView, the obj is a MissionDrone instance. getattr(obj, "mission_drones", None) returns None because MissionDrone doesn't have a mission_drones reverse relation. The fallback getattr(obj, "operator_id", None) == request.user.id does the right thing. However, for MissionOutcomeView, obj is a Mission, so mission_drones.filter(operator=request.user).exists() fires an extra query that is not select_related'd. The queryset in MissionOutcomeView uses Mission.objects.with_related() which only selects commander and created_by. Add .prefetch_related("mission_drones") to the queryset, or at least be aware that this is one query per permission check.

missions/serializers.py
MaksymKorchynskyi, line 111: MissionSerializer.Meta.fields does not include incident_notes. The field is writable through MissionOutcomeSerializer and stored in the DB, but it is invisible on GET /api/missions/ and GET /api/missions/{id}/. Add "incident_notes" to the MissionSerializer.fields list so consumers can read what was recorded.

missions/services.py
```diff
@@ -126,6 +134,161 @@ def assign_drone_to_mission(
...
+        )
+
+        update_drone(
+            drone=locked_drone,
+            drone_data={"status": target_drone_status},
+            user=action_user,
+            related_mission=locked_mission,
+            writeoff_reason=writeoff_reason,
+            written_off_at=timezone.localdate() if is_lost else None,
+        )
```
MaksymKorchynskyi, line 301: If an operator records condition_after = "ok" on a drone whose status is already ACTIVE, update_drone is still called. Inside update_drone, _update_instance_fields detects no change, so drone.save() is skipped, but a DroneStatusHistory row with from_status == to_status == "ACTIVE" is not created either (because of the old_status != drone.status guard in update_drone). So functionally it's harmless, but it's a wasted SELECT … FOR UPDATE + function call overhead. Consider short-circuiting when target_drone_status == locked_drone.status and the condition didn't change.

accounts/rbac.py
MaksymKorchynskyi, line 113: commander_permissions() does not include PERMISSION_MISSIONS_RECORD_OUTCOME or PERMISSION_MISSIONS_RECORD_CONDITION. Given that Commanders can already update mission status and authorise write-offs, it seems like an oversight that they cannot record outcomes. If this is intentional, add a comment explaining why. If not, add both permissions to commander_permissions().

missions/serializers.py
```diff
@@ -234,6 +245,83 @@ def validate(self, attrs):
...
+                        "Outcome can only be recorded for missions with "
+                        "status 'completed' or 'aborted'."
+                    ),
+                },
+            )
+        if "result" not in attrs:
+            raise serializers.ValidationError(
+                {"result": "This field is required."},
+            )
+        return attrs
```
MaksymKorchynskyi, line 305: Both MissionOutcomeSerializer.validate and record_mission_outcome check that the mission status is completed or aborted. The serializer validates against the unlocked instance, then the service re-validates against the locked row. This is conceptually correct (optimistic check → pessimistic check), but the service raises the same ValidationError text. If the service check fires, the user sees a 400 from the serializer layer — which is fine — but the duplicate makes maintenance harder. Consider removing the serializer-level check or explicitly documenting that it's a fast-path guard. The same pattern applies to MissionDroneConditionSerializer.validate and record_drone_condition.

missions/services.py
```diff
@@ -126,6 +134,161 @@ def assign_drone_to_mission(
...
+        if previous_condition == Condition.LOST and condition_after != Condition.LOST:
+            raise serializers.ValidationError(
+                {
+                    "condition_after": (
+                        "Cannot reverse a 'lost' condition: a writeoff record "
+                        "has been created and requires a manual reversal "
+                        "process."
+                    ),
+                },
+            )
```
mehalyna, line 264: The logic prevents changing from 'lost' to any other status, but doesn't prevent the reverse scenario where a drone already marked 'lost' in a previous mission could be set to 'lost' again, potentially creating duplicate writeoff records. Add a check to prevent recording 'lost' condition if the drone is already written off.

missions/services.py
```diff
@@ -126,6 +134,161 @@ def assign_drone_to_mission(
...
+            id=locked_assignment.drone_id,
+        )
+        previous_drone_status = locked_drone.status
+
+        is_lost = condition_after == Condition.LOST
+        writeoff_reason = (
+            "Mission outcome: drone marked as lost"
+            if is_lost
+            else f"Mission outcome: condition_after={condition_after}"
+        )
```
mehalyna: The writeoff_reason uses a generic message for non-lost conditions but still passes it to update_drone. This could create confusing audit trails when the drone isn't actually being written off. Only pass writeoff_reason when actually writing off the drone.
```
writeoff_kwargs = {}
if is_lost:
    writeoff_kwargs = {
        "writeoff_reason": "Mission outcome: drone marked as lost",
        "written_off_at": timezone.localdate()
    }
update_drone(drone=locked_drone, drone_data={"status": target_drone_status}, 
             user=action_user, related_mission=locked_mission, **writeoff_kwargs)
```

missions/permissions.py
```diff
@@ -15,6 +15,26 @@ def has_permission(self, request, view):
...
+    def has_object_permission(self, request, view, obj):
+        role_code = get_user_role_code(request.user)
+        if role_code == ADMIN_CODE:
+            return True
+        if role_code != OPERATOR_CODE:
+            return False
+        mission_drones = getattr(obj, "mission_drones", None)
+        if mission_drones is not None and hasattr(mission_drones, "filter"):
+            return mission_drones.filter(operator=request.user).exists()
+        return getattr(obj, "operator_id", None) == request.user.id
```
mehalyna: The permission check uses a generic getattr fallback that may return False positives. If 'obj' is neither a Mission nor a MissionDrone, the check getattr(obj, "operator_id", None) == request.user.id could pass incorrectly. Make the object type explicit and fail closed.

missions/serializers.py
```diff
@@ -234,6 +245,83 @@ def validate(self, attrs):
...
+        instance = self.instance
+        if instance and instance.status not in (Status.COMPLETED, Status.ABORTED):
+            raise serializers.ValidationError(
+                {
+                    "status": (
+                        "Outcome can only be recorded for missions with "
+                        "status 'completed' or 'aborted'."
+                    ),
+                },
+            )
```
mehalyna, line 281: The validation only checks if the mission status is 'completed' or 'aborted', but doesn't verify if an outcome has already been recorded. This allows operators to overwrite outcomes multiple times without restriction. Add validation to prevent duplicate outcome recording.

missions/services.py
```diff
@@ -126,6 +134,161 @@ def assign_drone_to_mission(
...
+        locked_mission.result = result
+
+        if notes is not None:
+            locked_mission.notes = notes
+            update_fields.append("notes")
+        if incident_notes is not None:
+            locked_mission.incident_notes = incident_notes
+            update_fields.append("incident_notes")
+
+        locked_mission.save(update_fields=update_fields)
```
mehalyna, line 186: The function updates notes and incident_notes conditionally based on whether they are None, but empty strings will still update the fields. This inconsistent handling could lead to unintentional overwrites. Use explicit checks for provided values.

missions/views.py
mehalyna, line 119: The queryset filters by mission_id from URL kwargs but doesn't validate that the assignment actually belongs to that mission before the permission check runs. A malicious user could potentially access assignments from other missions. Add mission validation in get_object.

```
def get_object(self):
    obj = super().get_object()
    if obj.mission_id != self.kwargs["pk"]:
        raise Http404("Assignment not found for this mission.")
    return obj
```

missions/models.py
```diff
@@ -77,6 +81,7 @@ class Mission(models.Model):
     ended_at = models.DateTimeField(null=True, blank=True)
 
     notes = models.TextField(blank=True)
+    incident_notes = models.TextField(blank=True)
```
mehalyna: The incident_notes field has no max_length constraint on TextField, which could allow unbounded data storage and potential DoS attacks. Add reasonable length validation to prevent abuse.

missions/serializers.py
mehalyna, line 305: The serializer requires 'result' field but doesn't provide clear validation for the relationship between result and incident_notes. A 'success' result with incident_notes or 'failure' without notes could be logically inconsistent. Add cross-field validation.

```
def validate(self, attrs):
    # ... existing validation ...
    if attrs.get("result") == Result.SUCCESS and attrs.get("incident_notes"):
        raise serializers.ValidationError({
            "incident_notes": "Incident notes should not be provided for successful missions."
        })
    return attrs
```

missions/services.py
```diff
@@ -126,6 +134,161 @@ def assign_drone_to_mission(
...
+        locked_assignment.condition_after = condition_after
+        update_fields = ["condition_after"]
+        if condition_description is not None:
+            locked_assignment.condition_description = condition_description
+            update_fields.append("condition_description")
+        locked_assignment.save(update_fields=update_fields)
+
+        locked_drone = Drone.objects.select_for_update().get(
+            id=locked_assignment.drone_id,
+        )
```
mehalyna: The function locks the drone separately after already locking the assignment and mission, creating a potential deadlock scenario if another transaction locks these resources in a different order. Lock all required resources in a consistent order at the start of the transaction.

```
with transaction.atomic():
    locked_assignment = MissionDrone.objects.select_for_update().select_related("mission").get(id=assignment.id)
    locked_mission, locked_drone = (
        Mission.objects.select_for_update().filter(id=locked_assignment.mission_id) |
        Drone.objects.select_for_update().filter(id=locked_assignment.drone_id)
    )
```

accounts/rbac.py
olenakshyk: Recording mission outcome / drone condition is intentionally excluded: those are post-mission reports by the assigned Operator (with Admin override). Commanders oversee - they assign, change status, and authorize write-offs - but do not file the operator's after-action report.

missions/serializers.py
olenakshyk: I would disagree with the first part: even during successful missions, incidents can occur (temporary loss of signal, minor collisions with the surface, near-misses that should be recorded). Prohibiting this approach would lead to the concealment of precisely those reports that the incident tracking system is designed to record.

missions/services.py
olenakshyk: The SELECT … FOR UPDATE on the drone is acquired earlier in record_drone_condition (canonical lock order Mission → Drone → MissionDrone), not inside update_drone — so it's already held by the time we get here and isn't avoidable by short-circuiting.
What's left on the no-change path is just _update_instance_fields returning [], plus a couple of skipped if blocks — no save, no WriteOffRecord, no DroneStatusHistory, no extra query. I'd rather keep the uniform "propagate condition → drone state via update_drone" flow than add a branch for a few microseconds.

missions/views.py
olenakshyk: The queryset already enforces this — get_queryset() filters by mission_id=self.kwargs["pk"], and DRF's get_object() then does get(pk=assignment_id) on that filtered queryset. The final SQL is WHERE mission_id = <pk> AND id = <assignment_id>, so an assignment from a different mission won't match and DRF raises 404 before has_object_permission ever runs. The proposed obj.mission_id != self.kwargs["pk"] check can never be True since the queryset already guarantees that equality. Same pattern is used in MissionAssignmentDetailView

mehalyna (approved):

MaksymKorchynskyi (approved):

WWINMWWITW (approved):

PR #73, opened 2026-06-05
Implement mission media artifacts upload 

MaksymKorchynskyi:
**What was done:**
- Created a new Django app `media` and registered it in `INSTALLED_APPS`
- Created `MissionArtifact` model in `media/models.py`
- Added database migration with dependencies on `missions` and `drones` migrations
- Configured `MEDIA_URL`, `MEDIA_ROOT`, and env-configurable `ARTIFACT_MAX_FILE_SIZE_MB` in `config/settings.py`
- Implemented `POST /api/missions/{id}/artifacts/` to upload files, save to local storage, and auto-fill metadata
- Implemented `GET /api/missions/{id}/artifacts/` to list all artifacts for a mission
- Implemented `GET /api/missions/{id}/artifacts/{artifact_id}/` to retrieve artifact metadata
- Implemented `DELETE /api/missions/{id}/artifacts/{artifact_id}/` to delete artifact and clean up the file
- Added file type validation for video (`mp4`, `avi`, `mov`), image (`jpg`, `jpeg`, `png`), and data (`csv`, `json`)
- Added file size limit validation via `ARTIFACT_MAX_FILE_SIZE_MB`
- Added validation that the mission exists before saving (404 if not found)
- Added RBAC permission checks using `PERMISSION_MEDIA_UPLOAD` and `PERMISSION_MEDIA_VIEW`, and added `PERMISSION_MEDIA_UPLOAD` to `dispatcher_permissions()` in `accounts/rbac.py`
- Added audit log entries on upload (`action='artifact_uploaded'`) and delete (`action='artifact_deleted'`) via the existing `AuditLog` model
- Created Django admin (`media/admin.py`) for the `MissionArtifact` model for easier management


media/models.py
```diff
@@ -0,0 +1,77 @@
...
+
+ALL_ALLOWED_EXTENSIONS = [ext for exts in ALLOWED_EXTENSIONS.values() for ext in exts]
+
+EXTENSION_TO_FILE_TYPE = {
+    ext: file_type for file_type, exts in ALLOWED_EXTENSIONS.items() for ext in exts
+}
+
+
+def artifact_upload_path(instance, filename):
+    ext = os.path.splitext(filename)[1].lower()
```
mehalyna, line 48: The file extension is extracted but never validated against allowed extensions. Add validation before generating the safe filename to prevent unauthorized file types.

media/models.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+
+    file = models.FileField(upload_to=artifact_upload_path)
+    file_type = models.CharField(
+        max_length=10,
+        choices=FileType.choices,
+    )
```
mehalyna, line 104: The file_type field is not automatically determined from the uploaded file extension. Add a clean method or save override to auto-populate it to prevent data inconsistency.

media/models.py
```diff
@@ -0,0 +1,77 @@
...
+    original_filename = models.CharField(max_length=255)
+    file_size = models.PositiveBigIntegerField(
+        help_text="File size in bytes",
+    )
```
mehalyna, line 108: The file_size field is defined but there's no logic to populate it automatically. Add a save override to set file_size from the uploaded file object.

media/models.py
```diff
@@ -0,0 +1,77 @@
...
+
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+
+    file = models.FileField(upload_to=artifact_upload_path)
+    file_type = models.CharField(
+        max_length=10,
+        choices=FileType.choices,
+    )
+    original_filename = models.CharField(max_length=255)
```
mehalyna, line 105: The original_filename field is not automatically populated from the uploaded file. Add logic to capture the original filename before the upload path transformation.

media/models.py
```diff
@@ -0,0 +1,77 @@
...
+        settings.AUTH_USER_MODEL,
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="uploaded_artifacts",
+    )
+
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+
+    file = models.FileField(upload_to=artifact_upload_path)
```
mehalyna: The file field lacks validation for maximum file size and allowed extensions. Add validators to enforce constraints at the model level.

media/models.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    uploaded_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="uploaded_artifacts",
+    )
+
+    title = models.CharField(max_length=255)
```
mehalyna, line 87: The title field has no constraints on minimum length or format. Add validation to prevent empty or whitespace-only titles.

media/models.py
```diff
@@ -0,0 +1,77 @@
...
+    VIDEO = "video", "Video"
+    IMAGE = "image", "Image"
+    DATA = "data", "Data"
+
+
+ALLOWED_EXTENSIONS = {
+    FileType.VIDEO: [".mp4", ".avi", ".mov"],
+    FileType.IMAGE: [".jpg", ".jpeg", ".png"],
+    FileType.DATA: [".csv", ".json"],
+}
```
mehalyna: The allowed extensions list is hardcoded and may need frequent updates. Consider moving this to Django settings or a database configuration for easier maintenance without code changes.

media/serializers.py
```diff
@@ -0,0 +1,72 @@
...
+            "description",
+            "file",
+            "file_type",
+            "original_filename",
+            "file_size",
+            "storage_backend",
+            "captured_at",
+            "uploaded_at",
+        ]
+        read_only_fields = fields
```
mehalyna: All fields including writable ones like mission, title, and description are marked as read-only, making this serializer unusable for updates.

media/serializers.py
```diff
@@ -0,0 +1,72 @@
...
+                f"{settings.ARTIFACT_MAX_FILE_SIZE_MB} MB limit."
+            )
+
+        return file
+
+    def validate(self, attrs):
+        file = attrs.get("file")
+        if file:
+            ext = os.path.splitext(file.name)[1].lower()
+            attrs["file_type"] = EXTENSION_TO_FILE_TYPE[ext]
```
mehalyna: The extension is extracted again after already being validated in validate_file, causing redundant parsing. Store the file_type during validate_file and reuse it in the validate method.

media/serializers.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def validate_title(self, value):
+        stripped = (value or "").strip()
+        if not stripped:
+            raise serializers.ValidationError("Title is required.")
+        return stripped
+
+    def validate_file(self, file):
+        if file.size is None:
+            raise serializers.ValidationError("Cannot determine file size.")
```
mehalyna: Checking if file.size is None is insufficient for detecting empty files. Add validation to reject zero-byte files which are likely corrupted or invalid.

media/services.py
```diff
@@ -0,0 +1,93 @@
...
+                    "title": title,
+                },
+            )
+    except Exception:
+        if saved_path:
+            try:
+                default_storage.delete(saved_path)
+            except Exception:
+                logger.exception("Failed to clean up orphaned file %s", saved_path)
+        raise
```
mehalyna, line 58: The bare except Exception clause catches all exceptions including database transaction errors, but the file cleanup logic assumes the file was saved when it may not have been. Narrow the exception handling to only cleanup after successful file save.

media/services.py
```diff
@@ -0,0 +1,93 @@
...
+            target_model="MissionArtifact",
+            target_id=artifact_id,
+            changes={
+                "mission_id": mission_id,
+                "file_type": artifact.file_type,
+                "original_filename": original_filename,
+                "title": artifact.title,
+            },
+        )
+        artifact.delete()
```
mehalyna, line 82: The file is deleted from storage after the database transaction commits, creating a window where the database record is gone but the file still exists. Move file deletion inside the transaction or handle it before the transaction to maintain consistency.

media/services.py
```diff
@@ -0,0 +1,93 @@
...
+            artifact = MissionArtifact.objects.create(
+                mission=mission,
+                uploaded_by=uploaded_by,
+                title=title,
+                description=description,
+                file=file,
+                file_type=file_type,
+                original_filename=original_filename,
+                file_size=file_size,
+                storage_backend="local",
```
mehalyna: The storage_backend is hardcoded to "local" regardless of the actual Django storage backend configuration. Detect the storage backend dynamically from default_storage or accept it as a parameter.

media/services.py
```diff
@@ -0,0 +1,93 @@
...
+                "file_type": artifact.file_type,
+                "original_filename": original_filename,
+                "title": artifact.title,
+            },
+        )
+        artifact.delete()
+
+    if file_name:
+        try:
+            default_storage.delete(file_name)
```
mehalyna: File deletion happens outside the transaction but there's no retry mechanism or orphaned file tracking if deletion fails. Consider implementing a cleanup job or storing failed deletions for later retry.

media/services.py
MaksymKorchynskyi: The reason I kept the exception block a bit broad is that Django saves the file to storage during pre_save (before the SQL insert). If create() fails because of a database error, the file will still be left on the disk. Catching these errors lets us clean up the orphaned file if the transaction rolls back, and the default_storage.exists() check ensures we only try to clean up if the file was actually saved

media/services.py
MaksymKorchynskyi: I considered moving the file deletion inside the transaction, but since file storage operations aren't transactional, if the DB transaction fails later on, the file would be permanently deleted while the record remains. Using transaction.on_commit felt like a safer way to make sure the file is only deleted if the transaction successfully completes

mehalyna (approved):

mehalyna (approved):

olenakshyk (approved):

WWINMWWITW (approved):

Issue #29, opened 2026-05-05
As a user, I can monitor and view the status and lifecycle history of any drone


mehalyna:
**User Story:**
As a user, I want to view real-time status (active, in mission, under repair, etc.) and the full lifecycle history for any drone, so I can see its usage, incidents, downtime, and significant events.

**Technical tasks:**
- Define allowed drone statuses
- Create DroneStatusHistory model: check table 'drone_status_history'
- Create migration for status history
- Automatically create history record on status change
- Add status history to drone detail response
- Show previous status, new status, changed by, reason, date
- Add tests for status history creation

**Acceptance Criteria:**
- Each drone record includes current status and history log (assignments, repairs, incidents, write-offs, etc.)
- Visual indicators show if a drone is in active mission, downtime, being repaired, or written off.
- History timeline is clearly presented and auditable from the UI.

PR #77, opened 2026-06-09
Implement drone lifecycle status history

yana-donda:
## Summary

Implemented drone lifecycle status tracking for issue #29.

## Changes

* Added `IN_MISSION` drone status.
* Added status history and visual status fields to drone API responses.
* Added `status_change_reason` support for drone updates.
* Updated mission lifecycle logic:

  * drones move to `IN_MISSION` when a mission starts;
  * drones return to `ACTIVE` when a mission is completed/aborted;
  * damaged drones are not overwritten back to `ACTIVE`.
* Removed duplicated drone status update logic from mission status view.
* Added tests for drone history and mission lifecycle behavior.


drones/serializers.py
```diff
@@ -145,19 +147,46 @@ class Meta:
...
             "related_writeoff",
             "created_at",
         )
         read_only_fields = fields
 
+    def get_changed_by_display(self, obj):
+        user = obj.changed_by
+
+        if not user:
+            return None
```
mehalyna: The method returns None when user is missing, but frontend typically expects empty string for consistency. Return empty string to avoid null-handling issues in the UI.

missions/serializers.py
```diff
@@ -374,6 +376,46 @@ def validate_status(self, value):
...
+                    if assignment.drone.status == Drone.STATUS_IN_MISSION:
+                        update_drone(
+                            drone=assignment.drone,
+                            drone_data={"status": Drone.STATUS_ACTIVE},
+                            user=user,
+                            related_mission=instance,
+                            status_change_reason="Mission finished",
+                        )
+
+        return instance
```
mehalyna, line 448: The method updates drone status during mission lifecycle transitions but doesn't handle transaction rollback failures. Wrap the entire method in a try-except to catch IntegrityError and provide meaningful error messages.

missions/serializers.py
```diff
@@ -374,6 +376,46 @@ def validate_status(self, value):
...
+        request = self.context.get("request")
+        user = getattr(request, "user", None)
+
+        with transaction.atomic():
+            instance.status = new_status
+            instance.save(update_fields=["status", "updated_at"])
+
+            assignments = instance.mission_drones.select_related("drone")
+
+            if old_status == Status.PLANNED and new_status == Status.ACTIVE:
```
mehalyna: The loop iterates through all assignments, but only checks if drone status is ACTIVE before changing to IN_MISSION. A drone could have status MAINTENANCE or DAMAGED at mission start, leading to silent skip. Log or validate all drone statuses before mission activation.

torii044 (approved):

mehalyna (approved):

WWINMWWITW (approved):

PR #81, opened 2026-06-09
implement bulk CSV import and export for drone inventory

M1ro1:
### 1. CSV Export (`GET /api/drones/export/`)
- Implemented a memory-efficient export feature using `StreamingHttpResponse` and Python generators (`EchoBuffer`).
- Integrated with the existing `DroneFilter`, allowing users to export data with applied filters and search queries (e.g., `?status=ACTIVE`).
- The exported file contains the following columns: ID, Serial Number, Inventory Number, Name, Model, Classification, Status, Military Unit, and Created At.

### 2. CSV Import (`POST /api/drones/import/`)
- Created an endpoint to accept `.csv` file uploads.
- Implemented robust row-level validation:
  - Validates the presence of required columns.
  - Skips rows with missing required fields or unknown `Military Unit` names.
  - Validates duplicate `Serial Number`s against the database.
- Implemented an error-aggregation mechanism: invalid rows are skipped, valid rows are saved, and a detailed summary (success count + row-level error array) is returned with a `200 OK` status (or `400 Bad Request` if the file structure is entirely invalid).
- *Note:* Added a temporary stub (`dummy_spec_data`) to satisfy the `DroneSpec` database constraints during import, as technical specifications are not provided in the basic logistics CSV. 

### 3. Testing & Refactoring
- Added comprehensive unit tests for both import and export endpoints (verifying pagination bypass, format validation, and row-level error tracking).
- Extracted `EchoBuffer` into a shared utility to maintain DRY principles.


anastasiiahumeniuk (approved):

drones/services.py
```diff
@@ -193,3 +198,143 @@ def update_drone(
...
+                    drone.created_at.strftime("%Y-%m-%d %H:%M:%S")
+                    if drone.created_at
+                    else ""
+                ),
+            ]
+        )
+
+
+def import_drones_csv(drones_csv_file, user=None):
+    decoded_file = drones_csv_file.read().decode("utf-8")
```
yana-donda: Could we handle `UnicodeDecodeError` here and return a validation-style error instead of letting the endpoint fail with a potential 500? The serializer only checks the .csv extension, but the uploaded file can still be invalid or use a different encoding.

drones/services.py
```diff
@@ -193,3 +198,143 @@ def update_drone(
...
+        serial_number = row.get("Serial Number", "").strip()
+        inventory_number = row.get("Inventory Number", "").strip()
+        name = row.get("Name", "").strip()
+        drone_model = row.get("Model", "").strip()
+        military_unit_name = row.get("Military Unit", "").strip()
+        acquired_at = row.get("Acquired At", "").strip()
+
+        if not all(
+            [serial_number, inventory_number, name, drone_model, military_unit_name]
+        ):
```
yana-donda, line 312: `acquired_at` is listed as a required column and is also documented as part of the expected import format, but empty row values for it are not validated here. Could we include `acquired_at` in this required field check and also validate the date format explicitly before passing it into create_drone_with_spec?

drones/services.py
```diff
@@ -193,3 +198,143 @@ def update_drone(
...
+            errors.append(
+                {"row": row_num, "error": "Missing one or more required fields."}
+            )
+            continue
+
+        if Drone.objects.filter(serial_number=serial_number).exists():
```
yana-donda, line 318: We already validate duplicate `serial_number`, but `inventory_number` is also unique on the `Drone` model. Could we add a similar explicit check for duplicate `inventory_number`?

drones/services.py
```diff
@@ -193,3 +198,143 @@ def update_drone(
...
+            errors.append(
+                {
+                    "row": row_num,
+                    "error": f"Drone with serial number "
+                    f"'{serial_number}' already exists.",
+                }
+            )
+            continue
```
mehalyna, line 326: The code checks for duplicate serial numbers but doesn't handle race conditions in concurrent imports. Use select_for_update() or database-level unique constraints to prevent duplicates during concurrent uploads.

drones/services.py
mehalyna, line 1: The import function `import_drones_csv` is not wrapped in a transaction, so partial imports will leave the database in an inconsistent state if an error occurs mid-processing. Wrap the entire import logic in @transaction.atomic or use transaction.atomic() context manager.

drones/views.py
```diff
@@ -48,3 +56,59 @@ def get_serializer_class(self):
...
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+
+        response = StreamingHttpResponse(
+            generate_drones_csv(queryset), content_type="text/csv"
+        )
+
+        response["Content-Disposition"] = 'attachment; filename="drones_export.csv"'
+
+        return response
```
mehalyna, line 98: The export view filters and streams all matching drones without a limit, which could cause memory/performance issues with large datasets. Add a maximum export limit similar to the audit log export (MAX_EXPORT_LIMIT).

drones/tests.py
```diff
@@ -737,3 +740,207 @@ def test_filter_by_typical_range_km_gte(self):
...
+                    "2026-05-11",
+                ],
+            ]
+        )
+
+        response = self.client.post(
+            self.import_url, {"file": csv_file}, format="multipart"
+        )
+
+        print(response.data["errors"])
```
mehalyna: A debug print statement should be removed.

mehalyna (approved):

yana-donda (approved):

PR #87, opened 2026-06-11
Drone comparison 

WWINMWWITW:
Side-by-Side Comparison: Created the DroneComparisonView to allow seamless analysis of hardware, performance, and firmware metrics across multiple drones.

Smart Highlighting: Added dynamic backend calculations to identify and visually highlight both the best (maximum) and worst (minimum) values for key performance metrics.

Export & Print Ready: Integrated a built-in CSV export feature (export_to_csv) and added native print-to-PDF support using window.print().

Secure RBAC Integration: Implemented the CanCompareDrones permission check within dispatch, explicitly denying access to the viewer role to meet business requirements without altering the shared rbac.py file.


drones/permissions.py
```diff
@@ -6,11 +6,38 @@
...
+        role_code = getattr(user_role_obj, "code", None)
+        if not role_code:
+            return False
+
+        if role_code == VIEWER_CODE:
+            return False
+
+        allowed_permissions = ROLE_PERMISSION_MATRIX.get(role_code, set())
+
+        return PERMISSION_SPECIFICATIONS_VIEW in allowed_permissions
```
mehalyna: The permission logic duplicates RBAC functionality already implemented in HasRBACPermission and hardcodes the VIEWER_CODE check. Use the existing HasRBACPermission class with PERMISSION_SPECIFICATIONS_VIEW instead of reimplementing the logic.

drones/views.py
mehalyna, line 1: 1. The min/max calculation logic is duplicated six times with nearly identical list comprehensions. Extract this into a helper method to reduce code duplication and improve maintainability.
2. Computing min/max values in the view layer with multiple database hits is inefficient for large querysets. Use Django's aggregation functions (Min/Max) to compute these values in a single database query.

drones/views.py
```diff
@@ -14,6 +18,181 @@
...
+
+    def get(self, request, *args, **kwargs):
+        ids_param = request.GET.get("ids", "")
+        drone_ids = self._parse_drone_ids(ids_param)
+
+        if request.GET.get("export") == "csv":
+            if len(drone_ids) > self.MAX_COMPARE_COUNT:
+                return HttpResponseBadRequest(
+                    f"Export failed. You can compare and export a \
+                     maximum of {self.MAX_COMPARE_COUNT} drones."
```
mehalyna: The message contains a line continuation backslash that creates awkward whitespace in the output. Use parentheses for implicit line continuation or a single-line f-string instead.

drones/views.py
```diff
@@ -14,6 +18,181 @@
...
+                    drone.classification,
+                    drone.status,
+                    getattr(spec, "firmware_version", "N/A"),
+                    getattr(spec, "max_speed_kmh", "N/A"),
+                    getattr(spec, "max_flight_time_min", "N/A"),
+                    getattr(spec, "payload_capacity_g", "N/A"),
+                ]
+            )
+
+        return response
```
mehalyna, line 178: The CSV export doesn't handle None values properly - getattr() returns "N/A" string but actual None values from the database will appear as empty strings. Use a consistent default value or explicit None handling.

drones/permissions.py
WWINMWWITW: I agree that hardcoding the VIEWER_CODE check wasn't ideal. However, I couldn't use the existing HasRBACPermission with PERMISSION_SPECIFICATIONS_VIEW because that permission grants access to the viewer role, which violates the acceptance criteria (restricting the feature to technical, operations, and dispatch staff only). To fix this properly, I have introduced a new dedicated permission in rbac.py specifically for drone comparisons and integrated it with HasRBACPermission

M1ro1 (approved):

mehalyna (approved):

yana-donda (approved):

PR #88, opened 2026-06-14
Task#65 component replacement logging

anastasiiahumeniuk:
### **Overview**
This pull request introduces comprehensive component replacement tracking for the Mission Control System. It adds a new `ComponentReplacement` model and associated APIs to record, filter, and export drone component maintenance history.

### **Key Features**

**1. Component Replacement Model**
- Tracks hardware replacements on drones (motors, batteries, cameras, frames, propellers, flight controllers, video transmitters, wiring, firmware, and other components)
- Records old and new serial numbers for accountability and traceability
- Stores replacement reason and timestamp
- Automatically tracks the technician who performed the replacement
- Includes comprehensive timestamps (created_at, updated_at)

**2. API Endpoints**
- **List & Create** (`/repairs/replacements/`) — View all replacements and create new ones
- **Detail View** (`/repairs/replacements/<id>/`) — Retrieve full replacement details
- **Export** (`/repairs/replacements/export/`) — Generate CSV reports with filtering support

**3. Advanced Filtering & Ordering**
- Filter by drone, component type, technician (replaced_by), and date range
- Support for start_date and end_date range queries
- Ordering by replaced_at, created_at, and component_type
- Pagination for large datasets

**4. Data Integrity & Validation**
- Prevents deletion of drones with active component replacements (PROTECT constraint)
- Validates that replacement dates cannot be in the future
- Requires component name when recording "OTHER" component types
- Automatic trimming of whitespace in text fields
- Server-side enforcement: replaced_by cannot be spoofed by client

**5. Admin Interface**
- Comprehensive admin panel for managing replacements
- List display with key fields (ID, drone, component type, serial numbers, technician, date)
- Advanced filtering and search capabilities
- Read-only access control (deletion disabled)
- Automatic technician assignment from authenticated user

**6. Export Functionality**
- CSV export with streaming for large datasets (up to 10,000 records)
- Includes drone details (ID, serial number) alongside replacement info
- Supports the same filtering options as the API

### **Technical Implementation**
- Django model with database indexes on frequently queried fields (drone+date, component_type+date, technician+date)
- DjangoFilterBackend for flexible filtering
- Two serializer variants: full detail and lightweight list views
- Factory methods for testing with realistic data
- Comprehensive test coverage (50+ test cases) including CRUD operations, validation, authentication, filtering, and CSV export

### **Security & Permissions**
- RBAC integration with RepairPermission checks
- Read-only access for viewers, full CRUD for admins
- Technician field automatically set from request context—cannot be manually assigned


repairs/views.py
```diff
@@ -1,12 +1,25 @@
...
+    ComponentReplacementListSerializer,
+    ComponentReplacementSerializer,
+    DefectReportListSerializer,
+    DefectReportSerializer,
+)
+
+
+class Echo:
+    def write(self, value):
+        return value
```
M1ro1: We have exactly the same function in common/utils.py, so I think it is better to import it from there

repairs/views.py
```diff
@@ -35,3 +48,88 @@ class DefectDetailView(generics.RetrieveAPIView):
...
+class ComponentReplacementExportView(generics.GenericAPIView):
+    permission_classes = [RepairPermission]
+    filter_backends = (DjangoFilterBackend,)
+    filterset_class = ComponentReplacementFilter
+
+    def get_queryset(self):
+        return ComponentReplacement.objects.select_related("drone", "replaced_by")
+
+    def get(self, request, *args, **kwargs):
+        max_export_limit = 10000
```
M1ro1: Right now this variable is hardcoded, but I think it is better to get MAX_EXPORT_LIMIT value from settings.py using getattr

from django.conf import settings

max_export_limit = getattr(settings, "MAX_EXPORT_LIMIT", 10000)

repairs/models.py
```diff
@@ -74,3 +89,71 @@ def __str__(self) -> str:
...
+                fields=["replaced_by", "-replaced_at"],
+                name="repl_user_replaced_idx",
+            ),
+        ]
+
+    def clean(self):
+        errors = {}
+
+        if self.replaced_at and self.replaced_at > timezone.now():
+            errors["replaced_at"] = "replaced_at cannot be in the future."
```
mehalyna: The validation logic checks self.replaced_at after it may be None, which will cause an AttributeError. Add a None check before comparing with timezone.now().

repairs/serializers.py
```diff
@@ -62,3 +62,85 @@ class Meta:
...
+        component_type = attrs.get("component_type")
+        component_name = attrs.get("component_name", "")
+
+        if component_type == ComponentType.OTHER and not component_name:
+            raise serializers.ValidationError(
+                {"component_name": ["Component name is required for OTHER."]}
+            )
+
+        if component_type != ComponentType.OTHER:
+            attrs["component_name"] = ""
```
mehalyna: Unconditionally clearing component_name for non-OTHER types will overwrite user input during updates. Only clear it during creation or when explicitly changed.

repairs/views.py
```diff
@@ -35,3 +48,88 @@ class DefectDetailView(generics.RetrieveAPIView):
...
+                        replacement.new_serial_number,
+                        replacement.reason,
+                        replacement.replaced_at.strftime("%Y-%m-%d %H:%M:%S"),
+                        (
+                            replacement.replaced_by.username
+                            if replacement.replaced_by
+                            else "N/A"
+                        ),
+                    ]
+                )
```
mehalyna: Using queryset.iterator() on a sliced queryset may load all rows into memory before iterating, defeating the streaming purpose. Apply filters without slicing or iterate over the full filtered queryset with a limit inside the loop.

repairs/serializers.py
```diff
@@ -62,3 +62,85 @@ class Meta:
...
+            "reason",
+            "replaced_at",
+            "replaced_by",
+            "created_at",
+            "updated_at",
+        )
+        read_only_fields = ("id", "replaced_by", "created_at", "updated_at")
+
+    def validate_component_name(self, value):
+        return (value or "").strip()
```
mehalyna: Returning an empty string for blank component_name is inconsistent with the model field which allows blank=True. Return the stripped value or None to preserve database NULL semantics.

repairs/models.py
```diff
@@ -74,3 +89,71 @@ def __str__(self) -> str:
...
+
+        if (
+            self.component_type == ComponentType.OTHER
+            and not self.component_name.strip()
+        ):
+            errors["component_name"] = "Component name is required for OTHER."
```
mehalyna, line 148: The validation checks component_name.strip() but this will fail if component_name is None (allowed by blank=True). Add None-safe checking.

repairs/views.py
```diff
@@ -35,3 +48,88 @@ class DefectDetailView(generics.RetrieveAPIView):
...
+    permission_classes = [RepairPermission]
+    filter_backends = (DjangoFilterBackend,)
+    filterset_class = ComponentReplacementFilter
+
+    def get_queryset(self):
+        return ComponentReplacement.objects.select_related("drone", "replaced_by")
+
+    def get(self, request, *args, **kwargs):
+        max_export_limit = 10000
+        queryset = self.filter_queryset(self.get_queryset())[:max_export_limit]
```
mehalyna: The hard limit of 10000 records applied after filtering could lead to incomplete exports without user notification. Add a response header or logging to indicate when the limit is reached.

M1ro1 (approved):

olenakshyk (approved):

mehalyna (approved):

Issue #79, opened 2026-06-09
User Story: Document Reason for Write-Off


mehalyna:
**Scope:**
As an operator, I want to document the specific reason for a drone's write-off (loss, destruction, critical damage, etc.) so that historical context is available.

**Technical Tasks:**
- Update write-off event model to support multiple canonical reasons (loss/destruction/damage/other + notes).
- Refactor API/UI to allow reason selection and custom note entry.
- Validation for reason codes.
- Unit tests for input validation and correct reason tracking.

**Acceptance Criteria:**
- [ ] User can select from a list of canonical reasons or specify another reason.
- [ ] Reason is saved, associated with the write-off event, and visible from write-off records.
- [ ] Custom details/notes are supported.
- [ ] Data validation prevents empty reason fields.

PR #89, opened 2026-06-15
Task #79 Reason choices for write-off

olenakshyk:
## Document write-off reasons for drones

Operators can now record why a drone was written off using a canonical reason, with optional free-form notes, so historical context is preserved on each write-off event.

### Changes
- **Model** (`WriteOffRecord`): `reason` is now a canonical choice field — `LOSS`, `DESTRUCTION`, `DAMAGE`, or `OTHER` — instead of free text. Added a `reason_label` property, model-level `clean()` validation, and `full_clean()` on save.
- **API**: the write-off flow (`DroneUpdateSerializer`) accepts a `writeoff_reason` selected from the canonical list plus a `writeoff_reason_description` for custom notes. Write-off records now expose both the raw `reason` code and a human-readable `reason_label`.
- **Validation**:
  - Reason is required and must be one of the canonical codes (empty/invalid values are rejected).
  - When `OTHER` is selected, a custom description is required.
- **Status history**: change-history entries now use the human-readable reason label.
- **Admin**: added `reason` filter and `reason_description` search


drones/models.py
```diff
@@ -362,6 +384,36 @@ class WriteOffRecord(models.Model):
...
+        errors = {}
+
+        if not self.reason:
+            errors["reason"] = "A write-off reason is required."
+        elif self.reason not in dict(self.REASON_CHOICES):
+            valid_codes = ", ".join(code for code, _ in self.REASON_CHOICES)
+            errors["reason"] = (
+                f'"{self.reason}" is not a valid write-off reason. '
+                f"Valid reasons: {valid_codes}."
+            )
```
mehalyna: The validation logic checks if reason is in REASON_CHOICES dict after already checking if it's empty, but this is redundant since the field has choices constraint.

drones/serializers.py
```diff
@@ -311,6 +314,14 @@ def validate(self, attrs):
                     "sold, transferred, or written off."
                 )
 
+        reason = attrs.get("writeoff_reason")
+        description = attrs.get("writeoff_reason_description") or ""
+
+        if reason == WriteOffRecord.REASON_OTHER and not description.strip():
+            errors["writeoff_reason_description"] = (
+                "A custom description is required when the reason is 'Other'."
+            )
```
mehalyna: The validation for REASON_OTHER duplicates the model-level validation in WriteOffRecord.clean().

drones/models.py
```diff
@@ -362,6 +384,36 @@ class WriteOffRecord(models.Model):
...
+            errors["reason_description"] = (
+                "A custom description is required when the reason is 'Other'."
+            )
+
+        if errors:
+            raise ValidationError(errors)
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        super().save(*args, **kwargs)
```
mehalyna, line 405: Calling full_clean() on every save can cause issues with Django admin and bulk operations. Override save() only when force_insert=True or validate explicitly in forms/serializers instead.

drones/tests.py
```diff
@@ -691,6 +694,176 @@ def test_drone_spec_change_log_rejects_invalid_new_values(self):
...
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("writeoff_reason", response.data)
+        self.assertFalse(WriteOffRecord.objects.filter(drone=self.drone).exists())
+
+    def test_blank_reason_is_rejected(self):
+        response = self._write_off(writeoff_reason="")
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("writeoff_reason", response.data)
```
mehalyna, line 820: This test expects a 400 error for blank reason, but the field is not blank=True, so Django/DRF will reject it before serializer validation. The test passes for the wrong reason (field requirement, not custom validation).

drones/services.py
```diff
@@ -183,14 +183,18 @@ def update_drone(
         )
 
     if old_status != drone.status:
+        writeoff_reason_label = (
+            writeoff_record.reason_label if writeoff_record else writeoff_reason
+        )
```
mehalyna: The writeoff_reason_label fallback logic could fail if writeoff_record is None but writeoff_reason contains a code instead of a label. Consider mapping the code to label.

drones/models.py
```diff
@@ -336,11 +336,33 @@ def __str__(self) -> str:
...
+    REASON_DESTRUCTION = "DESTRUCTION"
+    REASON_DAMAGE = "DAMAGE"
+    REASON_OTHER = "OTHER"
+
+    REASON_CHOICES = [
+        (REASON_LOSS, "Loss"),
+        (REASON_DESTRUCTION, "Destruction"),
+        (REASON_DAMAGE, "Critical damage"),
+        (REASON_OTHER, "Other"),
+    ]
```
mehalyna: The reason constants and choices are defined at class level but could be better organized using TextChoices (Django 3.0+) for cleaner code and automatic label generation.

mehalyna (approved):

torii044 (approved):

yana-donda (approved):

PR #90, opened 2026-06-15
Task#80 Record Drone Write-Off Event

torii044:
added write-off records API and updated drone status logic


drones/services.py
```diff
@@ -207,3 +207,41 @@ def validate_drone_classification(drone_model, classification):
...
+
+    DroneStatusHistory.objects.create(
+        drone=drone,
+        from_status=old_status,
+        to_status=drone.status,
+        changed_by=user,
+        reason=f"Status changed from {old_status} to {drone.status}",
+        related_mission=related_mission,
+        related_writeoff=writeoff_record,
+    )
```
mehalyna, line 463: The status history reason uses a generic template string instead of the human-readable reason label from the writeoff record. Use writeoff_record.reason_label to maintain consistency with the pattern established in PR #89.

drones/services.py
```diff
@@ -207,3 +207,41 @@ def validate_drone_classification(drone_model, classification):
...
+
+    return writeoff_record
```
mehalyna, line 465: The function accepts related_mission parameter but doesn't validate if the mission exists or belongs to the drone. Add validation or remove the parameter if not needed for this endpoint.

drones/serializers.py
```diff
@@ -373,3 +374,37 @@ def validate(self, attrs):
...
+
+class WriteOffRecordCreateSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = WriteOffRecord
+        fields = (
+            "id",
+            "drone",
+            "reason",
+            "reason_description",
+            "related_mission_id",
```
mehalyna: The related_mission_id field is included but related_mission parameter isn't passed to create_writeoff_record service. Either pass it through or remove it from the serializer fields.

drones/tests.py
```diff
@@ -862,3 +862,72 @@ def test_update_drone_classification_to_not_allowed(self):
...
+        self.create_url = reverse("drones:write-off-create")
+        self.drone = DroneFactory()
+        self.admin_user = AdminUserFactory()
+        self.viewer = ViewerUserFactory()
+
+    def test_create_valid_write_off_record_as_admin(self):
+        self.client.force_authenticate(self.admin_user)
+        payload = {
+            "drone": self.drone.id,
+            "reason": "Write Off Reason",
```
mehalyna: The test uses a free-text reason "Write Off Reason" instead of one of the canonical reason choices added in PR #89. Update to use WriteOffRecord.REASON_LOSS or similar valid choices.

drones/serializers.py
```diff
@@ -373,3 +374,37 @@ def validate(self, attrs):
...
+        drone = attrs.get("drone")
+
+        if drone.status in Drone.INACTIVE_STATUSES:
+            raise ValidationError(
+                {
+                    "drone": (
+                        f"Cannot write off drone with inactive status '{drone.status}'."
+                    )
+                }
+            )
```
mehalyna, line 526: The validation checks if drone status is in INACTIVE_STATUSES but should check if a writeoff record already exists to prevent duplicates. Add a check for existing writeoff records.

mehalyna (approved):

yana-donda (approved):

olenakshyk (approved):

PR #91, opened 2026-06-16
Task#64 repair management

yedynetsy:
### What was done:
* **Data Schema:** Added `status` field (`RepairStatus`) to the `DefectReport` model and created the `RepairEvent` model to maintain an immutable audit trail of repair actions.
* **Backend API:** * Implemented `POST /api/repairs/defects/{id}/update-status/` for status transitions with required comments.
  * Implemented `GET /api/repairs/defects/{id}/history/` to retrieve chronological repair history.
* **Business Logic & Security:** * Status updates are handled in a dedicated service layer (`services.py`) wrapped in `@transaction.atomic`.
  * Added strict workflow validation (e.g., preventing same-status updates, ensuring `VERIFIED` can only be reached from `FIXED`).
  * Enforced **RBAC**: Only Technicians, Commanders, and Admins can set `IN_PROGRESS` or `FIXED`; only Commanders and Admins can set `VERIFIED`.
* **Notifications:** Configured automated email notifications to the original defect reporter upon status changes.
* **UI:** Created an HTML/JS template for viewing defect details, submitting status updates, and dynamically fetching the repair history table without page reloads.
* **Testing:** Added comprehensive unit tests covering workflow logic, RBAC, permission edge cases, and email triggers. All 83 tests pass successfully.


repairs/migrations/0002_defectreport_status_repairevent.py
anastasiiahumeniuk, line 1: Migration conflict in `repairs`: there are two sibling `0002` migrations (`0002_componentreplacement` and `0002_defectreport_status_repairevent`). This should be resolved with a merge migration before merge.

repairs/views.py
anastasiiahumeniuk, line 1: I see the `defect_detail.html` template in the repo, but it doesn't appear to be connected to any Django view/URL yet. Is the UI page meant to be exposed in this PR as well?

repairs/services.py
```diff
@@ -29,6 +34,62 @@ def create_defect_report(
     )
 
 
+@transaction.atomic
+def update_defect_status(
+    *, defect: DefectReport, new_status: str, action_taken: str, user
+):
+    old_status = defect.status
+
+    if old_status == new_status:
```
MaksymKorchynskyi: The "defect" object is passed directly from the view and its status is checked without a database lock. This is a classic check-then-act race condition: concurrent requests could both read the same old_status and execute conflicting status transitions, creating inconsistent RepairEvent logs. To fix this, the view should pass the defect_id instead, and the service should lock the row using select_for_update() inside the @transaction.atomic block: defect = DefectReport.objects.select_related("drone", "reporter").select_for_update().get(pk=defect_id).

repairs/views.py
```diff
@@ -140,3 +180,9 @@ def generate_csv():
         response["X-Export-Limit"] = str(max_export_limit)
         response["X-Export-Truncated"] = str(export_truncated).lower()
         return response
+
+
+class DefectUIDetailView(LoginRequiredMixin, DetailView):
+    model = DefectReport
+    template_name = "repairs/defect_detail.html"
```
MaksymKorchynskyi: This generic DetailView uses only LoginRequiredMixin. It does not apply the RepairPermission RBAC check. Consequently, the established role-based access control system is bypassed, and any logged-in user can access the UI template regardless of their actual permissions. You need to verify that the user explicitly has the required permission to view repairs.

repairs/services.py
```diff
@@ -29,6 +35,65 @@ def create_defect_report(
...
+                    f"The status of the defect you reported"
+                    f" has changed to '{new_status}'.\n\n"
+                    f"Action taken: {action_taken}"
+                ),
+                from_email=settings.DEFAULT_FROM_EMAIL,
+                recipient_list=[defect.reporter.email],
+                fail_silently=True,
+            )
+
+    return event
```
mehalyna, line 97: The permission check for IN_PROGRESS and FIXED status updates does not validate that the user has the required role before attempting the update. If the role check fails, a PermissionDenied exception will be raised after the status is already saved to the database. Move all permission checks before the database update to ensure atomic validation.

repairs/views.py
```diff
@@ -47,6 +54,36 @@ class DefectDetailView(generics.RetrieveAPIView):
...
+
+        event = update_defect_status(
+            defect_id=pk,
+            new_status=serializer.validated_data["status"],
+            action_taken=serializer.validated_data["action_taken"],
+            user=request.user,
+        )
+
+        response_serializer = RepairEventSerializer(event)
+        return Response(response_serializer.data, status=status.HTTP_200_OK)
```
mehalyna, line 85: The post method passes request.user directly to update_defect_status() without checking if the user is authenticated. If an anonymous user reaches this endpoint, request.user will be an AnonymousUser object, causing the permission check to fail ambiguously. Enforce authentication in the view before invoking the service.

repairs/models.py
```diff
@@ -91,6 +104,32 @@ def __str__(self) -> str:
         )
 
 
+class RepairEvent(models.Model):
+    defect_report = models.ForeignKey(
+        DefectReport,
+        on_delete=models.CASCADE,
+        related_name="repair_events",
+    )
```
mehalyna, line 112: The foreign key uses on_delete=models.CASCADE, which will silently delete all repair events if a defect is deleted. This loses audit trail history. Use on_delete=models.PROTECT to prevent accidental deletion of defects with associated events.

repairs/services.py
```diff
@@ -29,6 +35,65 @@ def create_defect_report(
     )
 
 
+@transaction.atomic
+def update_defect_status(*, defect_id: int, new_status: str, action_taken: str, user):
+    defect = get_object_or_404(
+        DefectReport.objects.select_related("drone", "reporter").select_for_update(),
+        pk=defect_id,
+    )
```
mehalyna: The function uses @transaction.atomic but relies on get_object_or_404() which raises an exception outside the transaction context. If the defect is not found, the exception is raised without the transaction being rolled back properly. Use select_for_update() inside a transaction context to ensure proper locking and error handling.

MaksymKorchynskyi (approved):

mehalyna (approved):

anastasiiahumeniuk (approved):

PR #93, opened 2026-06-17
Task#83 mission video metadata

WWINMWWITW:
## Video Metadata for Mission Artifacts

**New `VideoMetadata` model** (`media/models.py`)
- FK to `Mission` (`CASCADE`) and `Drone` (`PROTECT`) with DB-level referential integrity
- Tracks `file_name`, `file_size`, `content_type`, `checksum`, `recorded_at`, `status`, `uploader`
- `FileExtensionValidator` + file size guard on upload; unique constraint on `file`
- Upload path: `missions/<id>/drones/<id>/<uuid>_<filename>`

**REST API** (`/api/media/videos/`)
- Full CRUD via `VideoMetadataViewSet` (routed through `urls_video.py`, separate from artifact routes)
- Filterable by `mission`, `drone`, `uploader`, `status`, `created_after/before`, `recorded_after/before`
- RBAC: `MediaViewPermission` / `MediaUploadPermission` / `MediaDeletePermission`
- Paginated responses via `StandardResultsSetPagination`

**File lifecycle** (`signals.py`)
- `post_delete` → deletes file from storage via `transaction.on_commit`
- `pre_save` → removes old file when replaced
- `perform_destroy` catches `ProtectedError` and returns 400 instead of 500

**`VideoUploadSerializer`** auto-fills `file_name`, `file_size`, `content_type`, `uploader`, `status=ready` on create; runs `ffprobe` to extract `duration_seconds`


M1ro1 (approved):

media/serializers.py
```diff
@@ -1,11 +1,116 @@
...
+        read_only_fields = [
+            "id",
+            "file_size",
+            "content_type",
+            "status",
+            "created_at",
+            "updated_at",
+            "uploader",
+            "duration_seconds",
+        ]
```
MaksymKorchynskyi, line 53: The `VideoMetadataViewSet` allows `update` and `partial_update` using the `VideoMetadataSerializer`, which permits updating the `file` field. However, the logic to recalculate `duration_seconds`, `file_size`, and `file_name` via `ffprobe` is entirely trapped inside `VideoUploadSerializer.create()`. If a user updates the video file via a `PATCH` request, the new file will be saved, but the metadata (`file_size`, `duration`, etc.) will still reflect the old file, permanently corrupting the data. Disable updates on the `file` field or move the `ffprobe` logic to a service where it can be reused for updates.

media/serializers.py
```diff
@@ -1,11 +1,116 @@
...
+        return obj.url
+
+
+class VideoUploadSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = VideoMetadata
+        fields = ["id", "mission", "drone", "file", "recorded_at", "checksum"]
+        read_only_fields = ["id"]
+
+    def create(self, validated_data):
```
MaksymKorchynskyi, line 65: `VideoUploadSerializer` accepts `mission` and `drone` foreign keys, but it never validates whether the selected `drone` was actually assigned to the specified `mission`. A user could link a video to a drone that never participated in that mission. Add validation to ensure a valid `MissionDrone` link exists before accepting the upload.

media/signals.py
```diff
@@ -0,0 +1,43 @@
...
+def delete_old_file_on_replace(sender, instance, **kwargs):
+    if not instance.pk:
+        return
+
+    try:
+        old_instance = VideoMetadata.objects.filter(pk=instance.pk).first()
+        if not old_instance:
+            return
+
+        old_file = old_instance.file
```
MaksymKorchynskyi: The `delete_old_file_on_replace` signal triggers a database query (`VideoMetadata.objects.filter`) on every save to fetch the old instance. Optimize this by tracking the original file state upon instance initialization to avoid an unnecessary DB hit.

media/views.py
```diff
@@ -89,3 +99,83 @@ def perform_destroy(self, instance):
...
+            permission_classes = [IsAuthenticated, MediaUploadPermission]
+        elif self.action in ["destroy", "update", "partial_update"]:
+            permission_classes = [IsAuthenticated, MediaDeletePermission]
+        else:
+            permission_classes = [IsAuthenticated]
+        return [permission() for permission in permission_classes]
+
+    def perform_create(self, serializer):
+        with transaction.atomic():
+            serializer.save()
```
MaksymKorchynskyi: In `VideoMetadataViewSet.perform_create`, `serializer.save()` is wrapped in `transaction.atomic()`. Since the serializer's `create` method spawns a blocking external subprocess (`ffprobe`), the database transaction remains open while waiting for the OS to finish reading the file. Under load, this will exhaust the database connection pool and cause the system to hang. Execute the `ffprobe` call outside of the database transaction.

media/serializers.py
MaksymKorchynskyi: Sorry, I just saw this will be added in the next PR #104

media/serializers.py
mehalyna, line 1: The VideoUploadSerializer.create() method calls ffprobe synchronously within the request-response cycle, which will block the entire request if ffprobe is slow or hangs. Use Celery or Django signals to extract duration asynchronously instead.

anastasiiahumeniuk:
https://github.com/ITA-Internship/Mission-Control-System/pull/104#discussion_r3459968880
The ffprobe subprocess call blocks the entire request and has no timeout, risking hangs on slow or unresponsive systems. Add a timeout parameter and consider moving duration extraction to an async task queue.

M1ro1 (approved):

mehalyna (approved):

MaksymKorchynskyi (approved):

PR #94, opened 2026-06-18
Implemented DevOps intergration for external storage providers

M1ro1:
Description:
Implemented DevOps integration for external storage providers (AWS S3, Azure Blob, GCP Storage, MinIO) based on environmental configurations.

Changes:

**Removed the single tests.py file and created a new tests/ package. Split the existing tests into separate, logically grouped files (e.g., test_models.py, test_services.py, etc.) to improve readability and maintainability.**

Added conditional storage backend configuration in settings.py driven by the STORAGE_PROVIDER env variable.

Enforced secure bucket/container policies: configured private object ACLs, disabled accidental file overwrites, and enabled Presigned URLs for secure access to mission artifacts.

Added comprehensive integration tests using mock to verify the upload and delete flows across all supported cloud providers without requiring real network calls.

Added configuration tests to ensure security parameters (like AWS_QUERYSTRING_AUTH) are correctly injected into the storage backend.


config/settings.py
mehalyna, line 1: Environment variables for cloud storage credentials are read without validation, risking runtime failures if variables are missing or malformed. Validate critical credentials at startup and raise ImproperlyConfigured immediately.

media/tests/test_storages.py
mehalyna, line 1: The security tests instantiate real storage backend classes without mocking, which may fail or hang if cloud credentials are invalid or network is unavailable during test runs. Mock the backend initialization and verify only the configuration values.

media/models.py
```diff
@@ -53,21 +52,7 @@ def artifact_upload_path(instance, filename):
 
 
 def _detect_storage_backend():
```
WWINMWWITW, line 75: This change breaks the self-verifying nature of the original implementation.
Previously, the function inspected the actual default_storage instance at runtime,
so storage_backend on the model always reflected what was truly active.
Now it just reads STORAGE_PROVIDER from settings — if there's ever a misconfiguration
where the settings value doesn't match the actual backend class in use, the field
will silently store wrong data with no error raised.

Consider keeping the introspection logic, or at minimum adding a startup check
(e.g. in AppConfig.ready()) that asserts the configured backend class matches
the STORAGE_PROVIDER value.

media/tests/test_storages.py
```diff
@@ -0,0 +1,183 @@
...
+from django.test import TestCase, TransactionTestCase, override_settings
+from storages.backends.azure_storage import AzureStorage
+from storages.backends.gcloud import GoogleCloudStorage
+from storages.backends.s3boto3 import S3Boto3Storage
+
+from media.services import delete_artifact, upload_artifact
+from missions.factories import MissionFactory, OperatorUserFactory
+
+
+class CloudStorageIntegrationTests(TransactionTestCase):
```
WWINMWWITW: These tests only assert that mock_save and mock_delete were called once,
but don't verify any database state. After upload_artifact, you should assert
that a MissionArtifact record was created with the correct storage_backend value.
After delete_artifact, assert the record no longer exists in the DB.
Without this, the tests confirm the mocks were hit but don't actually validate
the service contract.

Also, this class inherits from TransactionTestCase, but none of the tests
exercise on_commit behavior — everything is mocked. TransactionTestCase is
significantly slower than TestCase; it should only be used when you need
real transaction boundaries. I suggest switching to TestCase here.

PR #105, opened 2026-06-21
Preserve write-off history for auditing

yana-donda:
## Summary

- Added immutable write-off history records
- Added global and per-drone write-off history API endpoints
- Added simple write-off history report view
- Added role-based access control for write-off history
- Added tests for immutability, permissions, API, and report view

## Endpoints

- GET /api/drones/write-offs/
- GET /api/drones/<drone_id>/write-offs/
- GET /api/drones/write-offs/report/
- GET /api/drones/<drone_id>/write-offs/report/


drones/models.py
```diff
@@ -399,11 +423,22 @@ def clean(self):
...
 
     def save(self, *args, **kwargs):
-        if self._state.adding:
-            self.full_clean()
+        if self.pk and type(self).objects.filter(pk=self.pk).exists():
+            raise ValidationError(
+                "Write-off records are immutable and cannot be edited after creation."
+            )
 
+        self.full_clean()
```
mehalyna: The immutability check queries the database to detect updates, creating a race condition where concurrent requests can bypass the check. Use Django's update_fields parameter or check self._state.adding instead.

drones/models.py
```diff
@@ -399,11 +423,22 @@ def clean(self):
...
         super().save(*args, **kwargs)
 
+    def delete(self, *args, **kwargs):
+        raise ValidationError("Write-off records are immutable and cannot be deleted.")
+
+
+@receiver(pre_delete, sender=WriteOffRecord)
+def prevent_writeoff_record_delete(sender, instance, **kwargs):
+    raise ValidationError("Write-off records are immutable and cannot be deleted.")
+
```
mehalyna: Both the model's delete() method and a pre_delete signal raise identical exceptions, creating redundant protection and potential confusion about which layer prevents deletion. Remove the signal handler and rely on the model method.

drones/views.py
```diff
@@ -225,6 +227,99 @@ class DroneModelListCreateView(generics.ListCreateAPIView):
...
+
+        return user_has_permission(user, PERMISSION_WRITEOFF_VIEW)
+
+    def dispatch(self, request, *args, **kwargs):
+        if not self.has_writeoff_history_permission(request.user):
+            raise PermissionDenied(
+                "You do not have permission to view write-off history."
+            )
+
+        return super().dispatch(request, *args, **kwargs)
```
mehalyna, line 284: The permission check duplicates logic from WriteOffHistoryPermission class, violating DRY principle and risking divergence if one is updated. Use the permission class's _has_permission() method directly.

mehalyna (approved):

Issue #84, opened 2026-06-10
User Story: Secure RBAC-Protected Access to Mission Videos


mehalyna:
**User Story:**
As a user with assigned roles, I want to access mission videos only if my role and permissions allow, so that sensitive video content is protected and only available to authorized personnel.

**Technical Implementation:**
- Integrate the existing RBAC system into all video API endpoints.
- Define roles and permissions (e.g., admin, mission operator, auditor).
- Enforce access control for video upload, retrieval, deletion, and listing actions.
- Log all access and modification events for auditability.

**Acceptance Criteria:**
- Users without proper RBAC permissions cannot access mission videos.
- All video API endpoints enforce RBAC rules.
- Access and changes are logged with user and timestamp.
- Admins can review video access logs.

PR #106, opened 2026-06-22
Task#84 rbac for media

olenakshyk:
## Task #84 — RBAC for mission media

Enforces role-based access control and full auditability for mission media artifacts (video/image/data).

### Implemented
- **RBAC on every media endpoint** — upload / list / retrieve / delete gated by `media.upload`, `media.view`, `media.delete`.
- **Per-view access logging** — each retrieval of an individual artifact is recorded with the acting user, IP and timestamp.
- **New `MediaAuditLog` table** — dedicated audit log for media events (`view`, `upload`, `update`, `delete`) with FK to user, artifact and mission.
- **Dual-write for modifications** — upload/delete are written to **both** `MediaAuditLog` and `MissionAuditLog` inside a single `@transaction.atomic` block (best-effort, non-blocking for read/view logging).
- **Admin log review API** — `GET /api/media/audit-logs/`, read-only, guarded by the new `media.view_logs` permission (granted to `ADMIN` only), with filters: `action`, `user`, `mission`, `artifact`, `start_date`, `end_date`.


media/services.py
```diff
@@ -5,11 +5,66 @@
...
 logger = logging.getLogger(__name__)
 
 
+def _get_client_ip(request):
+    if request is None:
+        return None
+    x_forwarded_for = request.META.get("HTTP_X_FORWARDED_FOR")
+    if x_forwarded_for:
+        return x_forwarded_for.split(",")[-1].strip()
+    return request.META.get("REMOTE_ADDR")
```
mehalyna, line 26: The function extracts the last IP from X-Forwarded-For header, but this can be spoofed by malicious clients, logging incorrect IP addresses. Use the first untrusted IP (rightmost before trusted proxies) or configure Django's SECURE_PROXY_SSL_HEADER properly.

anastasiiahumeniuk (approved):

mehalyna (approved):

PR #108, opened 2026-06-26
Fix mission read RBAC bypass

yana-donda:
### Summary

Fixed RBAC bypass for mission read endpoints.

* Added `missions.view` permission check for mission list/detail access
* Prevented Technician and users without a role from reading mission data
* Added regression tests for unauthorized mission read access

### DIFF
--- a/missions/views.py
+++ b/missions/views.py
@@ -8,6 +8,7 @@ from accounts.rbac import (
     PERMISSION_MISSIONS_RECORD_CONDITION,
     PERMISSION_MISSIONS_RECORD_OUTCOME,
     PERMISSION_MISSIONS_UPDATE_STATUS,
+    PERMISSION_MISSIONS_VIEW,
 )
 
 from .models import Mission, MissionAuditLog, MissionDrone, Status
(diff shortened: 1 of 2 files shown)


missions/views.py
```diff
@@ -46,7 +51,11 @@ class MissionPagination(PageNumberPagination):
 
 class MissionListCreateView(generics.ListCreateAPIView):
     serializer_class = MissionSerializer
-    permission_classes = [permissions.IsAuthenticated, IsDispatcherOrAdmin]
+    permission_classes = [
+        permissions.IsAuthenticated,
+        MissionsViewRBAC,
+        IsDispatcherOrAdmin,
+    ]
```
mehalyna: The IsDispatcherOrAdmin permission check executes after MissionsViewRBAC, creating a permission escalation vulnerability. Users with missions.view permission can bypass dispatcher/admin checks on POST requests. Reorder permission classes to check role-based create permission first.

missions/views.py
```diff
@@ -148,7 +157,11 @@ def perform_update(self, serializer):
 
 class MissionAssignmentListCreateView(generics.ListCreateAPIView):
     serializer_class = MissionDroneSerializer
-    permission_classes = [permissions.IsAuthenticated, IsDispatcherOrAdmin]
+    permission_classes = [
```
mehalyna: The same as l.54

missions/views.py
mehalyna, line 1: The get_queryset method filters missions but permission checks may allow users to see filtered results despite failed create authorization. Consider adding permission checks within get_queryset or using get_queryset_for_list vs get_queryset_for_create pattern to separate concerns.

PR #109, opened 2026-06-26
Bugfix#102 high load bad request abuse

M1ro1:


accounts/views.py
```diff
@@ -133,17 +131,24 @@ class AuditLogViewSet(viewsets.ReadOnlyModelViewSet):
...
-                Q(actor=user) | Q(target_user=user)
-            ).select_related("actor", "target_user")
+            logs_as_actor = AuditLog.objects.select_related(
+                "actor", "target_user"
+            ).filter(actor=user)
+            logs_as_target_user = AuditLog.objects.select_related(
+                "actor", "target_user"
+            ).filter(target_user=user)
+
+            return logs_as_actor.union(logs_as_target_user).order_by("-created_at")
```
anastasiiahumeniuk: This optimization breaks filtered audit-log requests for users with PERMISSION_AUDIT_LOGS_VIEW_OWN.
get_queryset() returns logs_as_actor.union(logs_as_target_user), but both list and export later call self.filter_queryset(...), and Django does not support .filter() after union().
As a result, requests like GET /api/accounts/audit-log/?action_type=... or ?start_date=... can fail with NotSupportedError: Calling QuerySet.filter() after union() is not supported.
Could we keep this as a regular queryset, or move the optimization to a form that still allows downstream filtering?

missions/services.py
```diff
@@ -85,19 +85,19 @@ def assign_drone_to_mission(
                 },
             )
 
-        locked_drone = Drone.objects.select_for_update().get(id=drone.id)
-        if locked_drone.status != Drone.STATUS_ACTIVE:
+        current_drone = Drone.objects.get(id=drone.id)
```
anastasiiahumeniuk: Removing select_for_update() from Drone and User improves lock contention, but it also seems to reopen a race condition.
Right now only the Mission row is locked, so two concurrent transactions for two different missions can both read the same drone/operator as available, both pass _check_overlap(), and both create assignments.
There is no DB constraint here that protects the “no overlapping assignment” business rule across missions.
I think we need either a safer concurrency strategy here or a test that proves this case is still protected.

